PR #2, opened 2016-04-25
A simple config reader

Kingdorian:
I created a simple config reader as part of the backbone. This can be very usefull in the future when we have complicated parameters for the computer vision part or configuration details for the database. I also changed the directory structure so I suggest we merge this into develop so we can use a (more) correct directory structure from now on.


bvanwijk:
Looks very good, only one thing in Eclipse I get an error that the import of config can not be resolved.
The import is right now:  com.benine.backend.Config, shouldn't it be: main.java.com.benine.backend.Config

The same for the package declaration, it is now : benine.backend, I think it should be: test.java.benine.backend for example.

Kingdorian:
It should for not be main.java.com.... I believe.  But just com.benine.backend and com.benine.backendtest or something similar.
https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html
Also I see I acidentally created this pullrequest to merge into master I'll create a new one to request a merge into develop.

Kingdorian:
https://github.com/Kingdorian/Contextproject-BeNine/pull/3

PR #3, opened 2016-04-25
A simple config reader (correct merging develop)

Kingdorian:
I created a simple config reader as part of the backbone. This can be very usefull in the future when we have complicated parameters for the computer vision part or configuration details for the database. I also changed the directory structure so I suggest we merge this into develop so we can use a (more) correct directory structure from now on.


Kingdorian:
https://github.com/Kingdorian/Contextproject-BeNine/pull/2

backend/src/main/java/com/benine/backend/ConfigReader.java
```diff
@@ -0,0 +1,68 @@
...
+    // If the line only contains a comment return null
+    if(whiteSpaceRemoved.equals("")) {
+      return null;
+      // Check if line is a valid config line
+    } else if (wellFormed.matcher(whiteSpaceRemoved).matches()) {
+      // Check if the line contains data or is just a comment
+        String[] stringSplit = whiteSpaceRemoved.split("=");
+        return stringSplit;
+    } else {
+      throw new InvalidConfigFileException("Mallformed line: " + whiteSpaceRemoved);
```
jochem725: I think this should be Malformed instead of Mallformed.

jochem725:
I've added a small comment regarding a typo, the rest looks great! After the typo has been fixed, I'll agree with merging.

Kingdorian:
See issue #2 for the issue regarding the package name.  This is according to Oracle and maven standards I believe. I'll fix the typo.

bvanwijk:
I agree with merging.
To fix the error you can change the java build path in the properties of the Eclipse project.

Kingdorian:
If you use maven build does the error in eclips not automatically dissappear?

PR #4, opened 2016-04-26
Created a simple logger that logs to a file

Kingdorian:
It's a simple logger to log events that happen in the system. It writes the events to a file with a specified max size. If the file exceeds the limit it backs it up, clears it and starts over again. It keeps at most 1 backup.


jochem725:
Looks great! I agree with merging.

egedebruin:
Looks like a great simple logger, I also agree with merging.

NaomidR:
Very useful and easy-to-use logger.This can be merged

PR #5, opened 2016-04-26
Basic version of front-end server.

jochem725:
Features:
- Serves index.html file with a very simple Bootstrap template.
- Simple Logger
- Tests using the MochaJS framework.

Installing the server is easy:
First make sure NodeJS v4 is installed on your system, then inside the frontend folder run `npm install`. Now run `node app` to start the server. Tests can be run using `npm test`.


Kingdorian:
Looks good!

egedebruin:
Looks like a great start, I agree with merging.

NaomidR:
Looks good, a clear explanation of how to use it is also given!

PR #6, opened 2016-04-26
Update README.md

Kingdorian:
Updated the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,13 @@
 # Contextproject-BeNine
+
+
+## Frontend
+Installing the frontend:
+1. Make sure NodeJS v4 is installed on your system.
+2. Open a terminal inside the frontend directory.
+2. Run "npm instal".
+3. Run "node app" to start the server.
+Tests can be run by running npm test.
+
+## Backend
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,13 @@
-# Contextproject-BeNine
\ No newline at end of file
+# Contextproject-BeNine
+
+
+## Frontend
+Installing the frontend:
+1. Make sure NodeJS v4 is installed on your system.
+2. Open a terminal inside the frontend directory.
+2. Run "npm instal".
```
jochem725: This should be "npm install"

Kingdorian:
Whoops you're right I'll fix it.

egedebruin:
Is it correct that this should be merged in master?

Kingdorian:
@egedebruin Yes because its only a readme update. However, I see the formatting of the list is not correct.

jochem725:
There is an issue regarding dependencies on windows using the node install guide. I'll try to look into it as soon as possible so the readme can be correctly updated.

jochem725:
Since the problem is now fixed, this can be merged.

PR #7, opened 2016-04-27
Fix project structure

Kingdorian:
Fixed the project structure, we now have a super pom file. Running mvn test or mvn install in the root directory now applies to both (backend, cameramodule). Running maven in one of both projects is also still possible when running maven from the dir the project resides in. The jar of the camera module is now accessible in the backend project. 

This might require re-importing the project in your IDE (looking at you eclipse...).

Note: the backend depends on the jar created by the camera module so for building/running/testing the backend you should always run be an install/package in the root directory first. (I'll put this in the readme) Unfortunately maven cannot manage this AFAIK.


bvanwijk:
Looks good, I agree with merging

jochem725:
So if I understand this correctly, Maven now assumes there are two projects with separate test suites and runs them independent from each other?

Kingdorian:
Yes.

jochem725:
Ok, I think we can merge this.

PR #8, opened 2016-04-28
Node path fix

jochem725:
This should resolve all path related issues with NodeJS.


Kingdorian:
Someone with a windows pc should test this, could you please test this @bvanwijk, @NaomidR or @egedebruin ? If it works I agree with merging this.

bvanwijk:
Yes, I tested it with a windows pc and it works for me.
I agree with merging.

PR #9, opened 2016-04-28
Basicbackendhttp

Kingdorian:
Created a basic HTTP server with some endpoints, we can use this for the communication between the client and the backend.


bvanwijk:
Looks as a good start for the HTTP server, maybe you could add some more javadoc.

NaomidR:
I agree with Bryan about the javadoc, some big methods without any comments are present in the code. Beside that it looks good.

jochem725:
I do not agree with the fact that this pull request has been merged without resolving the comments. Merging this reduces the overall quality of the documentation.

PR #10, opened 2016-04-28
Camera communication

bvanwijk:
Created the IP camera communication module. You can now send commands to the camera's to control: zoom, movement, focus, iris and to get the stream link.
The commands are tested with a mocked server to verify the righ requests are made.


Kingdorian:
The camera interface looks really nice, the Ipcamera class is a bit to large I think. I think this can be merged but we should adress the size of the Ipcamera class in the future.

cameramodule/src/main/java/com/benine/ipcameracontrol/Ipcamera.java
```diff
@@ -0,0 +1,283 @@
...
+
+
+/**
+ * Class to communicate with an IP Camera.
+ * @author Bryan
+ */
+
+public class Ipcamera implements Camera {
+
+  String ipadres;
```
jochem725: Shouldn't this be ipaddress?

jochem725:
Looks nice, I've placed a small comment regarding a variable name. After this is fixed, I'll agree with merging.

NaomidR:
The IPcamera class should indeed become shorter in the next sprint. But for now I agree with merging.

egedebruin:
Looks good, is it correct that the .checkstyle files belong to the backend and camera module?

PR #11, opened 2016-04-28
Database

egedebruin:
Created a start for the databasemodule
There is an interface for the database to add, delete, update and get presets from the database.
I've tried to make it able to make a connection to a database, which I failed to do. I only made this able on my own computer.
There is also a sql-file in the database module which creates the database.


Kingdorian:
This should not be merged into master. Please close this and create a correct pullrequest (into develop).

NaomidR:
Agreed, does not belong in the master.

PR #15, opened 2016-04-28
Database to develop

egedebruin:
Created a start for the databasemodule
There is an interface for the database to add, delete, update and get presets from the database.
I've tried to make it able to make a connection to a database, which I failed to do. I only made this able on my own computer.
There is also a sql-file in the database module which creates the database.


Kingdorian:
Looks great! However the JavaDoc for the interfaces could be a bit more elaborate. But I think this can be merged.

bvanwijk:
Looks as a good basis for the database module, it can be improved next sprint.

PR #17, opened 2016-04-28
Sprint week 1

Kingdorian:


NaomidR:
I agree with merging

Kingdorian:
Should be fine.

jochem725:
Great! I agree with merging.

bvanwijk:
I think it can be merged

PR #20, opened 2016-04-30
Control button

NaomidR:
This would be the first version of the joystick. At the bottom of the page you can see a cube moving around in the position that you move the joystick. The movements can also be controlled with w,a,s,d. 

To install the virtualjoystick, run: bower install virtualjoystick.js

When resizing the window, the joystick sometimes moves to a wrong starting place. This still needs to be fixed.


frontend/public/index.html
```diff
@@ -184,11 +184,119 @@ <h5 >Preset 1</h5>
 	
 	</div>
 	
+	<!-- Additional information about the position of the cube. -->
+		<div id="CubePositionX" style="position:fixed; left:90%; bottom:4%; color:grey;"></div>
+		<div id="CubePositionY" style="position:fixed; left:90%; bottom:8%; color:grey;"> </div>
+		
+		<!--Scripts to be used for the joystick part-->
+		<script src="http://threejs.org/build/three.min.js"></script>
```
jochem725: I think we should look for these files as a bower dependency. The system is going to be used in a mostly offline environment, so the app cannot download these files in the fly. They need to be fetched before using them.

jochem725:
Looks fine, however I think there are some things that should be improved. Most of the javascript code in index.html can be included in a separate file so the file becomes more readable and the joystick is a separate open source project on GitHub, which we shouldn't just copy paste into the repository, but fetch as a dependency using bower.

jochem725:
After looking for the bower install command for virtual joystick, I've found a module which looks a bit cleaner, maybe we can replace the current joystick with this one. http://yoannmoinet.github.io/nipplejs/

frontend/public/index.html
NaomidR: I agree, that's a good point. I'll look into that.

NaomidR:
The code should indeed still be cleaned up a bit/ put into a different file. I am not sure about using the other module. The way the joystick looks on the page does indeed look better than the one I used. However, I did look into every line of code of the virtual joystick and added a clear explanation of the functions as well as changed things in the module. There was really clear documentation available for the virtual joystick and it works good with bootstrap/bower. I don't know how this is with the one you are suggesting?

jochem725:
@NaomidR Looks are not a problem, as they can easily be overwritten by CSS. With cleaner looks I was referring to the code quality and documentation of the module. For virtual joystick, I was unable to find documentation, there were only some code fragments available on the repository. NippleJS has a very clear documentation, and provides events in clear JSON. How does virtual joystick handle this?

Kingdorian:
Also another consideration is the licence under which the joystick module is released. If its a GPLv3 licence for example we cannot use the code for commercial use if we might do so in the future. Nipplejs is MIT I see which is good, I cannot really find info on the one @NaomidR is referring to.

jochem725:
@Kingdorian The module used by @NaomidR is also released under MIT.

bvanwijk:
Seems fine, I think we now have a basic version of a joystick.

jochem725:
We should be able to merge it now. Issue #47 contains a task required for the joystick to be able to actually control the camera's.

egedebruin:
Looks good indeed, I think we can merge it.

PR #21, opened 2016-05-01
Remove output during tests

bvanwijk:
The tests that use the Mockserver are stoped producing a lot of output during tests.
The added file configures the logging of the mockserver.


Kingdorian:
Seems fine.

jochem725:
Great, this will make the maven output much more readable.

egedebruin:
Seems like a great little fix.

PR #22, opened 2016-05-02
Restructured to get rid of external jars

Kingdorian:
I restructured so now the database and camera "modules" are just part of the main project.


bvanwijk:
Looks good, I agree with merging

jochem725:
Seems fine, I'll agree with merging.

NaomidR:
This looks like a better structure, I agree with merging.

PR #23, opened 2016-05-02
Fix failing test (because of charset) and ipcamera hotfix

Kingdorian:
Hotfix for the failing test of the max logsize (because of charset) on Bryan's (and possibly others) machine(s).

@bvanwijk

### DIFF
--- a/backend/src/main/java/com/benine/backend/LogWriter.java
+++ b/backend/src/main/java/com/benine/backend/LogWriter.java
@@ -146,6 +146,7 @@ public class LogWriter {
         }
         try {
           File backupFile = new File(logLocation + "-old.log");
+          writer.close();
           oldFile.renameTo(backupFile);
           logSize = 0;
           writer = new PrintWriter(new FileWriter(logLocation  + ".log"));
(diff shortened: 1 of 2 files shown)


bvanwijk:
It still fails on my machine.
The LogWriterTest.testMaxLogSize test gives a NoSuchFile error for the file logs\maxlogsizetestlog-old.log

Kingdorian:
@bvanwijk I'll look into this for me it creates the file if it does not yet exist, can you please make sure you have the right acces rights setup in the logs dir? Also does it pass if you manually create an empty file called: logs\maxlogsizetestlog-old.log?

jochem725:
It appears to be working correctly for me. Besides that, maybe you could add a tearDown for the test log files created, so they don't 'pollute' the log folder.

Kingdorian:
@jochem725 thats a good idea, I'll add that (late) tonight or tomorrow.

bvanwijk:
@Kingdorian when I create the file manually it still fails. After I have run maven the file I created manually is deleted.

bvanwijk:
@Kingdorian I think I fixed the problem. The rename operation in the hardWrite method dit not succeed, because the system saw it was still in use by another process.

jochem725:
The fix suggested by @bvanwijk is not affecting my machine.

Kingdorian:
@bvanwijk Silly windows.. it works on my machine so I think we can merge this if you agree.

bvanwijk:
Yes, I agree with merging.

PR #24, opened 2016-05-02
NodeJS Configuration file + Refactor

jochem725:
I've added a config.json file, which is generated automatically on first start.

It's persistent over multiple requires, so you can just call require and use it in all modules.
I've also added this pattern to the logger, to make it as easy as possible to use these modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,5 +54,8 @@ jspm_packages
 # Optional npm cache directory
 .npm
 
+# config files.
+*.json
+
 # Optional REPL history
 .node_repl_history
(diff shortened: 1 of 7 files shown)


bvanwijk:

NaomidR:
Nice that you made it persistent. I think that this can be merged.

egedebruin:
Good work, and also nice to implement it with the logger.

Kingdorian:
Looks good!

jochem725:
I'll merge it then!

Issue #12, opened 2016-04-28
Split IPcamera class into smaller classes


Kingdorian:
We should split the IPcamera class into smaller classes.

PR #26, opened 2016-05-02
Camera communication

bvanwijk:
There is now a cameracreator class which creates the camera objects.
If you want to communicate with the camera's you have to read the cameratype and additional information from the config and call with this information the camera creator.
This class gives back the right camera object which can be used to communicate with the camera's.

The IPCamera class is now split up in several classes for the different functions of a camera.
This way it is possible to just change the behaviour of one function of a camera without needing a new kind of camera object.

#12


backend/src/main/java/com/benine/backend/cameracontrol/FocusFunctions.java
```diff
@@ -0,0 +1,50 @@
+package com.benine.backend.cameracontrol;
+
+/**
+ * Interface for Focus functions of camera's.
+ * @author Bryan
+ */
+public interface FocusFunctions {
```
Kingdorian: Why not simply call this focus? Or focus controller. Functions is maybe a bit vague in my opinion.

backend/src/main/java/com/benine/backend/cameracontrol/IrisFunctions.java
```diff
@@ -0,0 +1,40 @@
+package com.benine.backend.cameracontrol;
+
+/**
+ * Interface for camera iris functions.
+ * @author Bryan
+ */
+public interface IrisFunctions {
```
Kingdorian: Same as for FocusFunctions.

jochem725:
@Kingdorian @bvanwijk I spoke with Bryan about this just a few minutes ago. I proposed a solution in which a Camera has a list of Attribute objects such as ZoomAttribute, etc. I think that would be cleaner.

backend/src/main/java/com/benine/backend/cameracontrol/ipcameracontrol/IpcameraZoom.java
```diff
@@ -0,0 +1,46 @@
...
+    String res = camera.sendCommand("%23GZ");
+    if (res.substring(0, 2).equals("gz")) {
+      return Integer.valueOf(res.substring(2), 16);
+    } else {
+      throw new IpcameraConnectionException("Getting the Zoom position of the camera failed.");
+    }
+  }
+  
+  /**
+   * position must be between 0 and 2730.
```
Kingdorian: missing jdoc for zpos

Kingdorian:
@jochem725 @bvanwijk Sounds like a good plan, maybe Operations would be a better name?

Apart from that, the code looks good apart from the couple of comments I already placed you might want to glance over the javadoc once more, there are some @params or @returns tags missing for example.

backend/src/main/java/com/benine/backend/cameracontrol/MovingCamera.java
```diff
@@ -0,0 +1,40 @@
...
+    * @throws CameraConnectionException when command can not be completed.
+    */
+  public abstract void move(int pan, int tilt) throws CameraConnectionException;
+
+  /**
+    * Get the absolute position of the camera at this moment.
+    * @return array with two values 0: Pan, 1: Tilt both in degrees.
+    * @throws CameraConnectionException when command can not be completed.
+    */
+  public abstract double[] getPosition() throws CameraConnectionException;
```
Kingdorian: Maybe we should change the array with 2 elements to a position object.

Kingdorian:
Apart from the little comment above I've got 3 other comments. 

 First, maybe you should incorporate the logger to log important actions that are performed/events that occur. 
The second thing is it might be better to create an abstract CameraFactory/Creator with implementations for each type of camera (For now only IPCamera) instead of the current more generic CameraCreator.  
The last thing is I think we might have to get the abstract class Camera to the main package or create a camera package with a sub package ipcamera.

jochem725:
I agree with @Kingdorian. As I am unable to run Checkstyle, PMD and FindBugs I assume you took this into account.

bvanwijk:
I think it can be merged now.

Kingdorian:
Seems like the issues I previously addressed are fixed now.

jochem725:
Seems like this can be merged indeed.

PR #28, opened 2016-05-03
Enhancelogger

Kingdorian:
I enhanced the logger. Instead of the logwriter the logger should now be used to log. It logs to a specified log file and to the console. Both functions can be disabled and reenabled during program execution.

We'll first need to merge the hotfix pushed to master into develop before this can be merged into develop.

https://github.com/Kingdorian/Contextproject-BeNine/pull/28

### DIFF
--- a/backend/src/main/java/com/benine/backend/LogWriter.java
+++ b/backend/src/main/java/com/benine/backend/LogWriter.java
@@ -88,7 +88,7 @@ public class LogWriter {
   /**
    * Writes a list of logEvents to the list.
    */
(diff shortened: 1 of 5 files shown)


bvanwijk:
Looks good, if the hotfix you mentioned is merged, I agree with merging this branch.

jochem725:
Seems fine, I am receiving an error regarding the path of the config file, but this only seems to affect my machine, so it can be resolved later.

egedebruin:
Looks good indeed, great enhancement

PR #29, opened 2016-05-03
Pull hotfix into develop

Kingdorian:
We pushed a hotfix to the filewriter problem into master this should also be merged into develop.


bvanwijk:
I agree

jochem725:
Looks fine, I think this can be merged because of the same reason @bvanwijk mentions.

PR #30, opened 2016-05-03
Json

Kingdorian:
Added toJSON methods for preset and camera (to be used when sending data to the frontend/client).

### DIFF
--- a/backend/src/main/java/com/benine/backend/ipcameracontrol/Camera.java
+++ b/backend/src/main/java/com/benine/backend/ipcameracontrol/Camera.java
@@ -141,7 +141,7 @@ public interface Camera {
   String getStreamLink();
 
   /**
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/benine/backend/ipcameracontrol/Camera.java
```diff
@@ -140,4 +140,10 @@ void moveTo(double pan, double tilt, int panSpeed, int tiltSpeed)
    */
   String getStreamLink();
 
+  /**
+   * Returns an JSON representation of this camera.
```
jochem725: Should be `returns _a_ JSON`

jochem725:
Except from a small typo, I agree with merging.

bvanwijk:
I agree with merging.

PR #35, opened 2016-05-04
Add final versions from ShareLatex

jochem725:


Kingdorian:
Looks good.

bvanwijk:
Agree with merging.

PR #36, opened 2016-05-04
Develop

jochem725:
This pull request merges develop into master, so we can tag the deliverables.


NaomidR:
I agree with merging!

bvanwijk:
I agree with merging to master.

PR #37, opened 2016-05-04
Fix loggertest carriage return windows and add different ports in test.

bvanwijk:
On my windows machine the logger test gave an error which is now fixed.
Can you confirm it still works for you?

I changed the port numbers of the mocked test camera's, so they can't interfer.


Kingdorian:
There is not much to say about this. Looks good.

egedebruin:
Good small fix.

jochem725:
Looks like a nice fix, this can be merged.

PR #38, opened 2016-05-04
Httpendpoints

Kingdorian:
I created the basic structure for the http endpoints. It's not the most clean java code I've ever written but Java is clearly not made for writing http servers.


backend/configs/main.conf
```diff
@@ -1 +1,5 @@
-ipadress=127.0.0.1
\ No newline at end of file
+ipadress=127001
```
jochem725: How exactly does an IP Address without . characters work?

Kingdorian: Its just a placeholder since we have no real camera's to use anyway. Since the fix for the configreader is nog yet in this branch we cannot use dots.

jochem725:
Seem fine, however test coverage is low and there are still some PMD and FindBugs errors that could be resolved.

Kingdorian:
You're right about PMD and Findbugs I'll take a look at those first thing tomorrow.

As for the test coverage I thought it would be really cumbersome to write (usefull) tests because we'd have to use a mockserver. However I just thought of a way I'd just have to mock the camera's, how did I not think of that...?  Tomorrow I'll look at adding some tests too.

bvanwijk:
Looks good after the additions, I agree with merging.

jochem725:
I think this can be merged.

backend/configs/main.conf
jochem725: Ok, then I assume this is not a problem.

PR #39, opened 2016-05-04
Fix maven-site malfunctioning

jochem725:
Maven site was broken in my setup.
It appears this was related to checkstyle.xml.

Is anyone affected by the same problem and does this pr solve the problem?

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -136,7 +136,7 @@
             <artifactId>maven-checkstyle-plugin</artifactId>
             <version>${checkstyle.plugin.version}</version>
             <configuration>
-               <configLocation>checkstyle.xml</configLocation>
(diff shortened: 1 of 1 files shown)


bvanwijk:
It works for me, is it right that the reportings of maven site for pmd, checkstyle and findbugs are in backend/target/site instead of target/site?

jochem725:
@bvanwijk Yes, this is correct for the current setup, since the master pom file is still in place, running maven site on each individual project. Maybe we should decide to get rid of the master pom file in the future.

Kingdorian:
Well we xd decided to keep te master pom file just in case.  But I think we won't need it anytime soon.

jochem725:
@Kingdorian I've added #40 so we can discuss about this later.

bvanwijk:
Alright I think we can merge it then

PR #41, opened 2016-05-05
Fix for wrong config path

jochem725:
I came across this bug which the test suite didn't catch.
Before anyone does a pull request. I suggest we are more careful with pull requests, because it seems reading code and running maven test is not always sufficient.


NaomidR:
Spotted well! I agree with merging this.

bvanwijk:
I also agree with merging. 
And a good point you mention.

PR #42, opened 2016-05-05
Fix clean up

bvanwijk:
I fixed all checkstyle warnings.
And have tried to remove the most of the PMD and findbugs warnings.
Some are still present, because some parts of the system are not yet finished.


egedebruin:
Great fix! I agree with merging.

jochem725:
Excellent! We should be careful not to introduce too much errors in the future.

Issue #43, opened 2016-05-05
Communication client - server


jochem725:
When we created the initial idea, we wanted to let the browser directly communicate with the server.
Since client and server share a different address, the browser poses this as a security threat.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

I propose we change the idea to let NodeJS communicate with the server, since this reduces a high amount of overhead.

bvanwijk:
In #53 this is fixed.

PR #44, opened 2016-05-05
Frontend presets

jochem725:
I've added automatic preset generation to the web-app. I've also put some work in automatic video preview generation. These elements can be automatically generated all having their own id.
If no camera or preset is present, it shows a placeholder image automatically generated by [HolderJS](http://imsky.github.io/holder/).

Please take a look at the jQuery code quality, since I think there is a lot of room for improvement.

Due to the problems encountered in issue #43, we need to change the way in which we fetch the presets from the database, so they are not yet automatically fetched.


bvanwijk:
I think it looks very good for now.
It gives a view of the interface.

egedebruin:
Looks good for the interface, great work!

PR #45, opened 2016-05-05
Database connection

egedebruin:
Everything to connect with the database including:
A MySQLDatabase class to connect to a MySQL Database and get, add, delete and update presets in the database.
Tests for the MySQLDatabase using a mock for the server.
In the Main the MySQLDatabase server will automatically start and create a new database if there isn't already one.


egedebruin:
Need to merge in develop.

PR #46, opened 2016-05-05
Database connection

egedebruin:
Everything to connect with the database including:
A MySQLDatabase class to connect to a MySQL Database and get, add, delete and update presets in the database.
Tests for the MySQLDatabase using a mock for the server.
In the Main the MySQLDatabase server will automatically start and create a new database if there isn't already one.


bvanwijk:
The code looks good and nice that you have fixed almost every checkstyle warning.
Is it possible to add some more tests for the databasePreset and MySQLDatabase class?

egedebruin:
I will add more tests for the databasePreset.
The MySQLDatabase class can't be tested much more, because there are limits in using the mock server.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -0,0 +1,247 @@
...
+      }
+      resultset.close();
+      statement.close();
+    } catch (SQLException e) {
+      e.printStackTrace();
+    }
+    return preset;
+  }
+
+  @Override
```
jochem725: This method contains duplicate code from the previous method, can this be resolved?

jochem725:
Should be fine except from a duplicate code issue. Also PMD and Findbugs have some very small issues which I think can be fixed.

egedebruin:
I changed the duplicate code and fixed a little config bug.

jochem725:
Great @egedebruin, I think this can be merged.

bvanwijk:
I agree with merging too.

PR #49, opened 2016-05-05
Develop

egedebruin:
I think this is sprint-ready.


jochem725:
I agree with @egedebruin. This can be merged into master.

bvanwijk:
Yes, it can be merged in master.

PR #50, opened 2016-05-06
Add sprint plan 3 and retrospective sprint 2.

bvanwijk:


jochem725:
I think we all agree this can be merged.

egedebruin:
Indeed, I think this can merged

PR #51, opened 2016-05-06
Add architecture design from Sharelatex.

bvanwijk:


jochem725:
I think we forgot to include this, great find @bvanwijk

PR #53, opened 2016-05-07
NodeJS bridge api

### LINKED ISSUE #43: Communication client - server

jochem725:
This provides a new api for the NodeJS front-end.

`/api/getinfo` returns info about the backend server in JSON format.
`/api/*` sends `*` to the backend server and returns the response to the client acting as a bridge between the client and the backend server.

And example of the use of the API is implemented by giving the Server Status icon in the client its functionality.

This should fix #43 by removing the communication client →  backend by adding the front-end server as a bridge: client →  front-end →  back-end

### DIFF
--- a/frontend/app.js
+++ b/frontend/app.js
@@ -2,6 +2,7 @@ var path = require('path');
 
 var logger = require('./modules/logger');
 var config = require('./modules/config');
+var api = require('./modules/api');
 
 var express = require('express');
 var app = express();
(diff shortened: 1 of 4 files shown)


bvanwijk:
Looks very good, nice that you have found a solution for this problem.
I agree with merging.

NaomidR:
Indeed a good solution, documented/commented well. I agree with merging.

Kingdorian:
I really like this solution, it's really clean.

PR #56, opened 2016-05-08
Responsive ui

jochem725:
I think this can be merged for now. I will continue to improve it during this sprint.
Please test this on multiple browsers, because web applications may be different across multiple devices.

The application should now only scroll in the preset area if the presets do not fit.
If any scrolling occurs outside of the preset area, please let me know.
 
I've currently hidden the control area on devices with a small screen size because small screen sizes make the app unusable.


NaomidR:
Looks good. You can also scroll the camera's but this is probably what you want as well. It works on chrome and internet explorer but doesn't seem to work on microsoft edge (a fairly new brower). It doesn't give any errors but doesn't load e.g. the scrolling part, the only working thing is the joystick.

bvanwijk:
Looks good, in chrome the preset area shows scrolling bars even when the presets fit without scrolling.

Kingdorian:
Looks good, works on chromium 49.0.2623.108. The only remark I have is the scroll bar is just the browsers default scroll bar, which does not fit nice with the style of the rest of the page.

NaomidR:
I do agree that the scrolling bar doesn't really fit with the rest of the page.

jochem725:
@NaomidR Microsoft Edge seems to lack support for motion jpegs in the img tag. 

@bvanwijk, @Kingdorian It seems like I cannot hide the scroll bar on Chrome and Chromium. I'm not sure if this is a real issue, since the primary focus is a tablet application. The 404 error is fired because Chrome tries to download a favicon, which the application doesn't have. I can create one if you want though.

NaomidR:
I don't think that it is a problem that it doesn't work on Microsoft Edge because nobody uses this (I think...) and I indeed also don't think that the scroll bar is an issue as for we are focussing on tablets. The most important thing is that it works on a tablet.

Kingdorian:
@jochem725 We're focussing on tables however I don't think we should focus to much on apple hardware, we should also keep android in the back of our minds since this is the main browser of android. However the scrollbar not disappearing  is not a huge problem. I really hate the looks however  (in the context of the rest of the pabe).

jochem725:
@Kingdorian Indeed, but the scrollbar issue does not appear on touch devices, since both android and iOS hide the scrollbar per default. I'm still trying to hide it though.

Kingdorian:
@jochem725 I didn't know that, then leave it as is!

jochem725:
@Kingdorian I was unable to fix the scrollbar issue, but I've added icons.

Kingdorian:
Great!

Kingdorian:
@jochem725 seems like travis is stuck.

jochem725:
@Kingdorian I think Travis is running again.

PR #57, opened 2016-05-09
Remove master pom

Kingdorian:
I removed the master pom file and updated .travis.yml accordingly.


bvanwijk:
I agree with merging.

NaomidR:
Yes, looks good. I agree with merging.

jochem725:
Great! I think this will remove some confusions.

PR #58, opened 2016-05-09
Make buffer and printwriter logwriter volatile

Kingdorian:
This should make the Logger thread safe. There is not really a clean way to test this using Junit, I did manually test this however and ikt works.


jochem725:

bvanwijk:
Looks good, I agree with merging!

NaomidR:
I also agree with merging this, looks just fine.

Issue #59, opened 2016-05-10
Joystick sending more move commands than the camera can handle


bvanwijk:
The joystick in the UI sends on every move event a request to move to the backend.
In the backend these requests are send to the camera which can't handle too much requests in a row.
According to the documentation of the panasonic camera's they can only handle commands which are send with a gap of 130 ms between them.

bvanwijk:
fixed in #98

PR #60, opened 2016-05-10
Camera control ui

bvanwijk:
It is now possible to control the camera's in the UI.
The control buttons are only visible if the selected camera supports this options.
You can test the control of a camera with my own camera.
The UI now retrieves the camera's from the backend server, so you have to start this server first.

#59 need still to be fixed.


backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+        throw new IOException();
+      }
+      return doc.toString();
+    } catch (IOException e) {
+      e.printStackTrace();
+    } catch (SAXException e){
+      e.printStackTrace();
+    } catch (ParserConfigurationException e){
+      e.printStackTrace();
+    }
```
jochem725: I think this catch statements can be merged into a single one, like
`catch (IOException|SAXException|ParseConfigurationException e){`

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+    return null;
+  }
+
+  @Override
+  public int getZoomPosition() throws CameraConnectionException {
+    //Not supported
+    return 50;
```
jochem725: Wouldn't it be better to return a value like -1 if an action is not supported? So the UI always takes a default value on -1?

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+  @Override
+  public String toJSON() throws CameraConnectionException {
+    JSONObject json = new JSONObject();
+    json.put("id", Integer.valueOf(this.id));
+    try {
+      json.put("pan", new Double(getPosition().getPan()));
+      json.put("tilt", new Double(getPosition().getTilt()));
+      json.put("zoom", new Double(getZoomPosition()));
+      json.put("iris", 20);
+      json.put("focus", 80);
```
jochem725: Why does this return 20 and 80 for iris and focus?

jochem725:
I've added some comments, overall I think this piece of code needs a bit more comments to make things (like the values) more understandable.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
Kingdorian: I don't think that would be better. If it throws a cameraconnectionexception the httprequest simply returns a 400 error. In my opinion the UI should then determine what it should display to the user.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
Kingdorian: If it has no focus and iris you can simply emit adding those to the json.

backend/src/main/java/com/benine/backend/Main.java
```diff
@@ -40,9 +41,28 @@ public static void main(String[] args) {
...
+    //Temporary adding multiple test camera's
     SimpleCamera camera = new SimpleCamera();
-    camera.setStreamLink("tuincam.bt.tudelft.nl/mjpg/video.mjpg");
+    camera.setStreamLink("http://tuincam.bt.tudelft.nl/mjpg/video.mjpg");
     cameraController.addCamera(camera);
+    Camera camera2 = new IPCamera2("83.128.144.84:88");
+    cameraController.addCamera(camera2);
+    SimpleCamera camera3 = new SimpleCamera();
+    camera3.setStreamLink("http://131.180.123.51/cgi-bin/nph-zms?mode=jpeg&monitor=4&scale=100&maxfps=6&buffer=100");
+    cameraController.addCamera(camera3);
```
Kingdorian: Please get stuff like this from the config, it makes the code way more readable.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+    } catch (SAXException e){
+      e.printStackTrace();
+    } catch (ParserConfigurationException e){
+      e.printStackTrace();
+    }
+    return null;
+  }
+
+  @Override
+  public int getZoomPosition() throws CameraConnectionException {
```
Kingdorian: Why return a hardcoded value of 50? If the camera does not support zoom it should not implement zoomingcamera and thus you won't need to implement this method.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+      sendCommand("zoomOut");
+    } else {
+      sendCommand("zoomStop");
+    }
+  }
+
+  @Override
+  public void moveTo(Position pos, int panSpeed, int tiltSpeed)
+      throws CameraConnectionException {
+    //not supported
```
Kingdorian: Do something to indicate this action is not supported.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
```diff
@@ -0,0 +1,175 @@
...
+    } else if (tilt < 15) { 
+      sendCommand("ptzMoveDown");
+    } 
+    
+    
+      
+  }
+
+  @Override
+  public Position getPosition() throws CameraConnectionException {
```
Kingdorian: It would be better to do something here to indicate the action is not supported.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
bvanwijk: This is just for testing purposes if they are not available the buttons will not be visible in the UI

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
bvanwijk: This was just for testing purposes so you could test the slider for the zoom position will be placed on the right value. The IPCamera2 class is going to be removed as soon as we can use the camera's from Polycast.

Kingdorian:
You're indicating this class is just for test purposes. However I think we should adhere to our normal code quality "rules" if we ever want to keep using this class in the future we won't have to rewrite parts.

I think you could've used a (elaborate) mockito mock object to achieve what you tried to achieve by writing the ipcamera2 class with a lot less work.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera2.java
jochem725: I agree with @Kingdorian about the exception, that would be more clear than returning -1. 
While this class is just for testing, we may be able to apply it to other camera's.

frontend/public/css/style.css
```diff
@@ -49,6 +49,53 @@ nav .label {
     bottom: 50px;
 }
 
+/* slider styling */
+input[type=range] {
+    -webkit-appearance: none;
+    background-color: #848484;
+    width: 200px;
+    height:55px;
```
jochem725: There are some indentation issues here.

frontend/public/js/control.js
```diff
@@ -0,0 +1,139 @@
...
+			cameras[JSON.parse(obj.cameras[c]).id] = JSON.parse(obj.cameras[c]);
+		}
+	}).done(function() { 
+		
+		var place  = 1;
+		var camera_area = $("#camera_area");
+		
+		// show stream of every camera in the carousel.
+		for (var c in cameras) {
+			camera_div = camera_area.find('#camera_' + place);
```
jochem725: Those variables are not declared.

frontend/public/js/control.js
```diff
@@ -0,0 +1,139 @@
...
+function setCurrentCamera(id){
+	currentcamera = id;
+	// Show the current camera in the editing view.
+	$('#current_camera').find('img').attr("src", cameras[currentcamera].streamlink);
+	camera_title = $('#current_camera').find('.camera_title');
+	camera_title.find('#camera_title').text(cameras[currentcamera].id);
+	
+	//determine which elements of the UI to show
+	if (cameras[id].zoom == undefined) {
+		$('.zoomslider').hide();
```
jochem725: Maybe you could do these jQuery calls once and store them in a variable.

frontend/public/js/control.js
```diff
@@ -0,0 +1,139 @@
...
+			place++;
+		}
+	});
+}
+
+function setCurrentCamera(id){
+	currentcamera = id;
+	// Show the current camera in the editing view.
+	$('#current_camera').find('img').attr("src", cameras[currentcamera].streamlink);
+	camera_title = $('#current_camera').find('.camera_title');
```
jochem725: This variable is not declared.

jochem725:
I've added some comments to the JS, you can check this using http://jshint.com. Also works excellent with IntelliJ.

bvanwijk:
I have checked the JS with jshint.com and fixed the errors.
I replaced the IPCamera2 class with a test class which creates a mocked camera which can be used for testing.
The test class runs the server so you can use the UI to control this camera.
Because this class runs a server I have excluded it from the tests in maven.

jochem725:
@bvanwijk Great! But why exactly should it be excluded from the maven tests?

bvanwijk:
It isn't an automated test, it just starts a server and that's it.
Using this you can test the controls from the UI on my test camera.

Kingdorian:
Looks very good!

backend/src/test/java/com/benine/backend/CameraControlTest.java
```diff
@@ -0,0 +1,188 @@
...
+    cameraController.addCamera(camera3);
+  }
+  
+  /**
+   * This methods starts a server used to manually test the user interface.
+   * @throws InterruptedException when the thread is interrupted.
+   * @throws IOException when the server can not be created.
+   */
+  @Test
+  public void ManualTestUI() throws InterruptedException, IOException{
```
Kingdorian: When this is merged in develop and the httpUpdate is merged in develop, we'll have to change this to a HttpController which makes this a class (and especially this part) a lot cleaner.

bvanwijk: Yes, I agree with that.

PR #64, opened 2016-05-11
Node config validator

### LINKED ISSUE #54: Outdated NodeJS config file causes unresponsive app
An outdated NodeJS config file causes an unresponsive app.
A solution would be validation of this config file, and adding missing values so the program doesn't crash.

jochem725:
I've added a validator to the NodeJS config, so the program no longer crashes on an invalid configuration, but uses the default. Everything is logged in the log file.

This should fix #54

### DIFF
--- a/frontend/modules/config.js
+++ b/frontend/modules/config.js
@@ -21,7 +21,7 @@ var Config = function() {
 /**
  * Validates the config file keys. If one of the default keys is missing or there is a syntax error, it returns false.
  * @param config   Config object.
- * @returns {*}
+ * @returns {boolean}
  */
 Config.prototype.validate = function (config) {
     var default_keys = Object.keys(default_config);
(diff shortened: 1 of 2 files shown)


frontend/modules/config.js
```diff
@@ -19,6 +19,35 @@ var Config = function() {
 };
 
 /**
+ * Validates the config file keys. If one of the default keys is missing or there is a syntax error, it returns false.
+ * @param config   Config object.
+ * @returns {*}
```
bvanwijk: This returns a boolean right?

bvanwijk:
Looks good, nice that you have tested this very well.

frontend/modules/config.js
jochem725: Nice find, I've fixed it.

Kingdorian:
Looks great!

jochem725:
I'll merge it then.

PR #67, opened 2016-05-11
Presets

NaomidR:
Class that creates new presets. In the requestHandler a new method called responseMessage is added that is easier to use than respond(exchange,response).

### DIFF
--- a/backend/src/main/java/com/benine/backend/http/RecallPreset.java
+++ b/backend/src/main/java/com/benine/backend/http/RecallPreset.java
@@ -10,10 +10,7 @@ import com.benine.backend.camera.ipcameracontrol.IPCamera;
 import com.sun.net.httpserver.HttpExchange;
 
 import java.io.IOException;
-import java.util.Random;
 import java.util.jar.Attributes;
-
-
 
 public class RecallPreset extends RequestHandler {
   
@@ -39,14 +36,13 @@ public class RecallPreset extends RequestHandler {
       int presetID = Integer.parseInt(parsedURI.getValue("presetid"));
       Preset preset = Main.getDatabase().getPreset(cameraID,presetID);
       
-      IPCamera ipcamera =  (IPCamera)getCameraController().getCameraById(cameraID);
+      IPCamera ipcamera = (IPCamera)getCameraController().getCameraById(cameraID);
       
       movingCamera(ipcamera,preset);
-     
+      responseSuccess(exchange);
     } catch (MalformedURIException e) {
       responseFailure(exchange);
       Main.getLogger().log("Wrong URI", LogEvent.Type.CRITICAL);
-      return;
     } catch (CameraConnectionException e) {
       // TODO Auto-generated catch block
       e.printStackTrace();
@@ -60,14 +56,12 @@ public class RecallPreset extends RequestHandler {
    * @throws CameraConnectionException exception thrown when camera cannot connect.
    */
   public void movingCamera(IPCamera ipcamera, Preset preset) throws CameraConnectionException {
-    Position position = new Position(preset.getPan(),preset.getTilt());
-    ipcamera.moveTo(position, preset.getPanspeed() , preset.getTiltspeed());
+    Position position = new Position(preset.getPan(), preset.getTilt());
+    ipcamera.moveTo(position, preset.getPanspeed(), preset.getTiltspeed());
     ipcamera.zoomTo(preset.getZoom());
     ipcamera.moveFocus(preset.getFocus());
     ipcamera.setAutoFocusOn(preset.isAutofocus());
     ipcamera.setIrisPos(preset.getIris());
     ipcamera.setAutoIrisOn(preset.isAutoiris());
-    
-    
   }
 }
--- a/backend/src/test/java/com/benine/backend/http/RecallPresetTest.java
+++ b/backend/src/test/java/com/benine/backend/http/RecallPresetTest.java
@@ -1,8 +1,7 @@
 package com.benine.backend.http;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-
+import com.benine.backend.Main;
+import com.sun.net.httpserver.HttpExchange;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+          //Adding the new preset to the database
+          Main.getDatabase().addPreset(cameraID, randomInt, preset);
+          responseMessage(exchange, true);
+        } else {
+          responseMessage(exchange, false);
+        }
+      }
+    } catch (MalformedURIException e) {
+      responseMessage(exchange, false);
+      Main.getLogger().log("Wrong URI", LogEvent.Type.CRITICAL);;
```
jochem725: This `;` can be removed.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      return;
+    }   
+    
+  }
+  
+  /**
+   * @param ipCamera the ipCamera you want to get the position of.
+   * @return DatabasePreset preset.
+   * @throws CameraConnectionException thrown if camera isn't an IP Camera.
```
jochem725: The exception in this comment is not thrown.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      int zoom = ipCamera.getZoomPosition();
+      int pan = (int)ipCamera.getPosition().getPan();
+      int tilt = (int)ipCamera.getPosition().getTilt();
+      int focus = ipCamera.getFocusPos();
+      int iris = ipCamera.getIrisPos();
+      boolean autofocus = ipCamera.isAutoFocusOn();
+    
+      //Create new DatabasePreset and return it.
+      DatabasePreset preset = new DatabasePreset(pan,tilt,zoom,focus,iris,autofocus);
+      return preset;
```
jochem725: You can return new `DatabasePreset(...)` immediately, without an extra variable.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return null;
+  }
+  
+  /**
+   * 
+   * @param exchange the HttpExchange.
+   * @param correct boolean that is true if the exchange is successful. False otherwise.
+   */
+  public void responseMessage(HttpExchange exchange, boolean correct) {
```
jochem725: This is duplicate code, the same method can be found in the superclass `RequestHandler`

backend/src/test/java/com/benine/backend/http/PresetCreationHandlerTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.benine.backend.camera.CameraConnectionException;
+import com.benine.backend.camera.CameraController;
+import com.benine.backend.camera.Position;
+import com.benine.backend.camera.ipcameracontrol.IPCamera;
+import com.benine.backend.database.DatabasePreset;
+
+import junit.framework.Assert;
+
+@SuppressWarnings("deprecation")
```
jochem725: This deprecation warning can be removed by changing the import statement from `unit.framework.Assert` to `org.junit.Assert`.

backend/src/test/java/com/benine/backend/http/PresetCreationHandlerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    PresetCreationHandler handler = new PresetCreationHandler(controller);
+    IPCamera ipcamera = mock(IPCamera.class);
+    when(ipcamera.getFocusPos()).thenReturn(33);
+    when(ipcamera.getIrisPos()).thenReturn(50);
+    when(ipcamera.getPosition()).thenReturn(new Position(0, 0));
+    when(ipcamera.getZoomPosition()).thenReturn(100);  
+    when(ipcamera.isAutoFocusOn()).thenReturn(true);
+
+    DatabasePreset preset = new DatabasePreset(0,0,100,33,50,true);
+    Assert.assertEquals(preset.getFocus(), handler.getCameraPositions(ipcamera).getFocus());
```
jochem725: I don't think we should use multiple assert statements in a single test, instead make a parameterized test for each attribute that needs an Assert statement. A single assertion per test makes it more easy to find the cause of a failure.

backend/src/test/java/com/benine/backend/http/RequestHandlerTest.java
```diff
@@ -42,6 +42,29 @@ public final void testRespond() throws Exception {
     verify(out).close();
   }
 
+  @Test
```
jochem725: You could move the creation of a mock camera, handler and exchange to an `@before` method, reducing duplicate code.

jochem725:
Test coverage of the PresetCreationHandler is 41%, this could be higher. Also, check PMD and Checkstyle, they still have some warnings that could be resolved.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: Fixed

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: Fixed!

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: Fixed :)

backend/src/test/java/com/benine/backend/http/PresetCreationHandlerTest.java
NaomidR: Changed and warning deleted.

Kingdorian:
I'll take an in dept look at this later today. On first glance it looks okay but there are some sketchy parts (most of which Jochem already pointed out).

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: Indeed, I moved it there. I deleted it here.

backend/src/test/java/com/benine/backend/http/RequestHandlerTest.java
NaomidR: Fixed and changed in all tests.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    Attributes parsedURI;
+    try {
+      parsedURI = parseURI(exchange.getRequestURI().getQuery());
+        
+      int cameraID = Integer.parseInt(parsedURI.getValue("id"));
+      Camera camera =  getCameraController().getCameraById(cameraID);
+      IPCamera ipCamera = null;
+      
+      if (camera instanceof IPCamera) {
+        ipCamera = (IPCamera)camera;
```
Kingdorian: You can remove line 42 and change this line (line 45) to IPCamera ipCamera = ....

backend/src/main/java/com/benine/backend/http/RequestHandler.java
```diff
@@ -45,6 +45,22 @@ public Attributes parseURI(String uri) throws MalformedURIException {
   }
 
   /**
+   * Formats the response message for a success or failure.
+   * @param exchange the HttpExchange.
+   * @param correct boolean that is true if the exchange is successful. False otherwise.
+   */
+  public void responseMessage(HttpExchange exchange, boolean correct) {
```
Kingdorian: It would be cleaner to split this method in two methods in my opinion, i.e. respondSucces(HttpExchange.... ) and respondFailure(...)

backend/src/test/java/com/benine/backend/http/PresetCreationHandlerTest.java
```diff
@@ -0,0 +1,39 @@
...
+  public void testGetCameraPositions() throws CameraConnectionException {
+    CameraController controller = mock(CameraController.class);
+    PresetCreationHandler handler = new PresetCreationHandler(controller);
+    IPCamera ipcamera = mock(IPCamera.class);
+    when(ipcamera.getFocusPos()).thenReturn(33);
+    when(ipcamera.getIrisPos()).thenReturn(50);
+    when(ipcamera.getPosition()).thenReturn(new Position(0, 0));
+    when(ipcamera.getZoomPosition()).thenReturn(100);  
+    when(ipcamera.isAutoFocusOn()).thenReturn(true);
+
```
Kingdorian: As Jochem pointed out elsewhere, you can split this up into multiple (Prameterized) methods which will aid finding the problem if (one of the) the test(s) fail(s).

backend/src/test/java/com/benine/backend/http/RequestHandlerTest.java
Kingdorian: You should hover make sure at the end of every test method that out is emptied. This could otherwise result in quirky behaviour.

backend/src/main/java/com/benine/backend/http/RequestHandler.java
NaomidR: I don't really agree with this. This method is created to get rid of all the String response = ..... in the code everywhere. Now you can just call this method and say if it is successful (true) or failed (false).

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: Fixed

backend/src/main/java/com/benine/backend/http/RequestHandler.java
Kingdorian: Then you could make it to respond respondsucces if it succeeded and respondfailure if it did not.

Kingdorian: Now you essentially have one method with two different types of behaviour specifying which behaviour you want by the boolean as argument. I think its cleaner to devide this in two methods. However if you and the others don't agree with this. You can atleast just make it:
 ...{
    respond(exchange, response = "{\"succes\":\""+correct+"\"}";  
   }

NaomidR: I do indeed agree that that is cleaner code so I will change it to that.

jochem725: I agree with Dorian about the method splitting, it spreads the behavior more nicely.

jochem725:
@NaomidR Great that you've made the tests smaller. However, I think responseMessage indeed becomes more clear if it is split in respondSuccess and respondFailure. Because this makes its use clear immediately, without needing users of the method to look into documentation.

NaomidR:
@jochem725 @Kingdorian Okay, I can see why you want it that way, I'll change this first thing tomorrow.

backend/src/main/java/com/benine/backend/http/RequestHandler.java
NaomidR: It has been splitted into two methods.

PR #68, opened 2016-05-12
Add information about coverage tools and tests

jochem725:
This provides the TA with information about where to find the generated reports.


NaomidR:
This is useful for the TA, i agree with merging.

bvanwijk:
Seems fine, I agree with merging.

Kingdorian:
Great.

PR #69, opened 2016-05-12
Http update

Kingdorian:
I updated the http api to be more clear. There might need to be changed a couple of things after we merge this with Naomi since I saw she also made some good changes in the code in the http package.


backend/src/main/java/com/benine/backend/Main.java
```diff
@@ -57,27 +48,17 @@ public static void main(String[] args) {
...
-      server.start();
       while (true) {
         Thread.sleep(100);
       }
-    } catch (IOException e) {
-      logger.log("Unable to start server", LogEvent.Type.CRITICAL);
     } catch (InterruptedException e) {
-      logger.log("Unable to start server", LogEvent.Type.CRITICAL);
+      e.printStackTrace();
+      logger.log("Exception occured", LogEvent.Type.CRITICAL);
```
jochem725: I don't think this log message is clear enough. `Exception occurred` does not make clear what kind of exception was thrown.

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
```diff
@@ -29,18 +31,15 @@ public CameraInfoHandler(CameraController controller) {
    * @throws IOException When writing the response fails.
    */
   public void handle(HttpExchange exchange) throws IOException {
-    //TODO add logging stuff
+    getLogger().log("Got an http request with uri: " + exchange.getRequestURI(), LogEvent.Type.INFO);
     JSONObject jsonObj = new JSONObject();
     ArrayList<Camera> cameras = getCameraController().getCameras();
     JSONArray camerasJSON = new JSONArray();
     for (Camera cam : cameras) {
```
jochem725: Maybe the camerasJSON object can be returned by a separate method, to split handling from generating the JSON a bit more.

backend/src/main/java/com/benine/backend/http/HttpController.java
```diff
@@ -0,0 +1,86 @@
...
+    server.createContext("/camera/", new CameraInfoHandler(camController, logger));
+    for (Camera cam : camController.getCameras()) {
+      createHandlers(cam);
+    }
+  }
+  /**
+   * Creates the handlers for a certain camera.
+   * @param cam camera to create handlers for.
+   */
+  public void createHandlers(Camera cam) {
```
jochem725: Should this method be public? It is only used locally as far as I can see.

Kingdorian: Jep that slipped through, I'll fix it.

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
Kingdorian: Great idea I'll create a getCamerasJSON method in the cameraController even.

backend/src/main/java/com/benine/backend/Main.java
Kingdorian: Good point.

jochem725:
Looks great, I've added some comments. Also check PMD, Findbugs and Checkstyle, as they have some solvable warnings.

bvanwijk:
Looks good, if you have fixed Jochems comments than I agree with merging.

NaomidR:
This indeed looks good and I agree with merging after you have fixed Jochems comments.

Kingdorian:
The presets branch must be merged first.

Kingdorian:
It should be fine now, however we still have to merge Naomi's presets branch into develop.

jochem725:
Great, I'll agree with merging.

PR #72, opened 2016-05-12
Preset ui

bvanwijk:
The UI can now load the presets for every camera.
It can load the preset image from the backend.
The presets don't have an ID yet, because this is not yet implemented in the backend. 
The call to recall a preset in the backend has also to be added.
I think it would be good to merge this when theses things are available.


backend/src/main/java/com/benine/backend/Main.java
```diff
@@ -64,6 +67,7 @@ public static void main(String[] args) {
       server.createContext("/move", new MovingHandler(cameraController));
       server.createContext("/zoom", new ZoomingHandler(cameraController));
       server.createContext("/preset", new PresetHandler(cameraController));
+      server.createContext("/public", new FileHandler(cameraController));
```
Kingdorian: What does this public endpoint do? It seems like the images for the presets are stored here but the name is vague.

backend/src/main/java/com/benine/backend/http/FileHandler.java
```diff
@@ -0,0 +1,61 @@
...
+  public void handle(HttpExchange exchange) throws IOException {
+    URI uri = exchange.getRequestURI();
+    String path = uri.getPath();
+    File file = new File("." + path).getCanonicalFile();
+    //check if the requested file exists
+    if (!file.isFile()) {
+      respond(exchange, "{\"succes\":\"false\"}");
+    } else {
+      String mime = "text/html";
+      if (path.substring(path.length() - 4).equals(".jpg")) {
```
Kingdorian: I think it would be cleaner to make a get mime method that uses the String.split(".") method and uses the last part to check what the extention is. You could even create a "extention"->"meme" map for an even cleaner implementation because this style might grow into a forest of if statements or a switch statement.

backend/src/main/java/com/benine/backend/Main.java
bvanwijk: Good point I will change it to /images

backend/src/main/java/com/benine/backend/http/FileHandler.java
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      Headers header = exchange.getResponseHeaders();
+      header.set("Content-Type", mime);
+      exchange.sendResponseHeaders(200, 0);              
+
+      OutputStream os = exchange.getResponseBody();
+      FileInputStream fs = new FileInputStream(file);
+      final byte[] buffer = new byte[0x10000];
+      int count = 0;
```
Kingdorian: What is the reason you're not using a  BufferedInputStream?(https://docs.oracle.com/javase/7/docs/api/java/io/BufferedInputStream.html)

backend/src/main/java/com/benine/backend/Main.java
Kingdorian: Static would be a good name to since this endpoint seems to serve the static content.

bvanwijk:
@Kingdorian I have added a getMime method as you suggested and this method uses a map to get the right mime type.

Kingdorian:
Great changes! I think we can merge this.

jochem725:
Looks great, an improvement you could add would be to make only the selected preset orange, and the other dark again. Now you can select multiple presets at the same time.

jochem725:
Looks fine.

jochem725:
@bvanwijk The program crashes on startup, `Exception in thread "main" java.sql.SQLException: Column count doesn't match value count at row 1`

backend/src/main/java/com/benine/backend/Main.java
```diff
@@ -26,13 +22,15 @@
...
+  @SuppressWarnings("unused")
   private static HttpController httpController;
 
   /**
    * Main method of the program.
    * @param args command line arguments.
+   * @throws SQLException When sql statement failes.
    */
-  public static void main(String[] args) {
+  public static void main(String[] args) throws SQLException {
```
Kingdorian: I don't think this is good practice.

Kingdorian:
Apart from the comment I made it looks good, I think we can merge it.

bvanwijk:
I changed something in the database, so you must first delete the database so the system can create the new updated database.

bvanwijk:
I think this branch is now ready to be merged?

jochem725:
I think this can be merged. Next person to agree can merge this.

Kingdorian:
Looks good.

PR #73, opened 2016-05-12
Handling the presets in Backend

egedebruin:
I have added methods in the cameracontroller class to handle the presets of the cameras in the backend. You can add, get and update the presets, and also reset the presets in de database.

This ensures that the program keeps on working if the database crashes, because the presets are kept in the memory.

The databasepreset class is now a preset class in de backend package.
I have also added tests en fixed checkstyle issues.


jochem725:
I have some small remarks about MySQLDatabase. There are some checkstyle line length issues. Also FindBugs has some issues. It also appears there is duplicate code which should be removed.

backend/src/main/java/com/benine/backend/Preset.java
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Returns a JSON representation of this object.
+   *
+   * @return JSON representation of this object.
+   */
+  public String toJSON() {
+    JSONObject json = new JSONObject();
+
+    json.put("pan", new Integer(pan));
```
jochem725: Why is a new instance created here? Pan, tilt etc. should already have the correct types.

backend/src/main/java/com/benine/backend/Preset.java
```diff
@@ -0,0 +1,184 @@
...
+      return false;
+    }
+    if (panspeed != preset.panspeed) {
+      return false;
+    }
+    if (autoiris != preset.autoiris) {
+      return false;
+    }
+    
+    return autofocus == preset.autofocus;
```
jochem725: Maybe an additional if statement for autofocus, and a default `return true;` would make this more clear.

jochem725:
In general, I think we should try to reduce the size of pull requests. Due to merging other branches, this one has become quite large.

egedebruin:
I have fixed the minor issues and the MySQLdatabase class, with the majority of the findbug issues.

bvanwijk:
I fixed the last findbugs error, I now agree with merging.

PR #74, opened 2016-05-12
Add reference to backend API

Kingdorian:
I added a link to the api docs to the readme so it can be found more easlily.


bvanwijk:
Nice!

jochem725:
Excellent!

PR #75, opened 2016-05-12
Develop ui fix

jochem725:
This should resolve the UI issues on develop.


jochem725:
I've fixed the slider orientation on responsive layouts, it should at least look little bit better now. Please check if it looks correct.

bvanwijk:
I think it can be merged now, I fixed a small problem with disappearing buttons when not supported.

jochem725:
Great find @bvanwijk, I think we can merge it now.

egedebruin:
It also looks good for me.

PR #76, opened 2016-05-12
Fix slider orientation

jochem725:
I fixed the slider orientation, so large values are now back on top again.


bvanwijk:
Nice, it works normal again I agree with merging

egedebruin:
Good small fix, merging is okay for me.

NaomidR:
This can indeed be merged

Issue #78, opened 2016-05-13
Presets in Database removed


bvanwijk:
The presets stored in the database aren't available anymore on the next system startup.

bvanwijk:
Fixed in #92

PR #82, opened 2016-05-13
Fix presets removing

egedebruin:
#78 
The presets are not removed anymore when the system starts.


bvanwijk:
Seems like an easy fix, I agree with merging.

jochem725:
This fix can be merged.

PR #87, opened 2016-05-13
HttpController createContext test cases

jochem725:
I've added HttpController test cases, to make sure the API is not changed without a test failure.


backend/src/main/java/com/benine/backend/http/HttpController.java
```diff
@@ -67,7 +67,7 @@ private void createHandlers() {
    * Creates the handlers for a certain camera.
    * @param cam camera to create handlers for.
    */
-  private void createHandlers(Camera cam) {
+  public void createHandlers(Camera cam) {
```
Kingdorian: If you changed this to public just for testing purposes I think you should create a mock object and use that for testing, that way you can access private methods.

Kingdorian:
Looks good I think it can be merged after the little thing pointed out in the comment is fixed.

bvanwijk:
Good improvements, I agree with merging.

backend/src/main/java/com/benine/backend/http/HttpController.java
```diff
@@ -24,33 +24,46 @@
 
   /**
    * Constructor, creates a new HttpController object.
-   * @param address an internetsocet adress indicating the ports for the server to listen to.
+   * @param address an internetsocet address indicating the ports for the server to listen to.
    * @param logger the logger to use to log to.
    * @param camController the cameracontroller that contains
    *                      the camera's which this server interacts with.
    */
   public HttpController(InetSocketAddress address, Logger logger, CameraController camController) {
```
Kingdorian: We might want to change this from address to, String adress, int port.

Kingdorian:
Looks good.

jochem725:
@Kingdorian Good point about the address and the port, I've added it.

jochem725:
I've performed a manual merge, please check if it has been done correctly.

PR #88, opened 2016-05-13
Update architecture design and add camera UML

egedebruin:
I have updated the architecture design with a database description and added an UML for the camera.


jochem725:
I think this can be merged.

bvanwijk:
I agree with merging.

PR #90, opened 2016-05-13
Add sprint plan 4 and add retrospective sprint 3

bvanwijk:


jochem725:
Since we agreed on this on ShareLatex and during the meeting, I think this can be merged.

Issue #84, opened 2016-05-13
Remove SQL dump junk from console.


Kingdorian:
When running the tests a lot of junk is put onto the console by the databasecontroller this should be removed or put into a seperate dedicated file.

bvanwijk:
Fixed in  #92

Issue #89, opened 2016-05-13
Database spamming on startup


jochem725:
The database appears to be spamming the console on startup, it also looks like a new database is created on each start which shouldn't be the case if the database is persistent.

bvanwijk:
![2016-05-13 2](https://cloud.githubusercontent.com/assets/9803346/15258255/d177e706-194a-11e6-85a5-f944763da895.png)

bvanwijk:
fixed in  #92

PR #92, opened 2016-05-13
Spamming of the database is now to a log file

egedebruin:
The database will not dump text to the console anymore, but to a different log file.

#84


jochem725:
Looks fine @egedebruin, however I think this is not going to work on all operating systems due to the line separator. I think `File.separator` instead of `/` is the solution for this.

bvanwijk:
I agree with Jochem

Kingdorian:
Also a better name for the log would be database.log, or even database-[database_name].log.

NaomidR:
I agree with Jochem and Dorian. The rest looks good.

egedebruin:
I have fixed the things you commented on and made it able to use the database if the database is already present before the program start ups, fixing #89

bvanwijk:
I'm happy you fixed this, it works for me.
However there are some checkstyle and findbugs warnings introduced, I would like it if you can fix them.

bvanwijk:

PR #93, opened 2016-05-14
Change name Pos to Position 

NaomidR:
Changed the names IrisPos and FocusPos to IrisPosition and FocusPosition. Added checkstyle to gitignore as well.


bvanwijk:
Seems like a good change.
I think we should add the .checkstyle file to the gitignore, because it's just for the eclipse checkstyle plugin.

Kingdorian:
I think its a good change however small things like this can just be included as part of the refactoring when you touch code and don't need a separate pull request.

NaomidR:
Yes the checkstyle file should indeed not be included. I created a pull request for this because it resolves an open issue.

jochem725:
@NaomidR Looks fine, I think this can be merged if the Checkstyle file removed from git. @Kingdorian I think we indeed require pull requests for this, at least to make everyone aware of the change.

bvanwijk:
Seems fine, you can merge it.

NaomidR:
@Kingdorian The Checkstyle file is removed and added to gitignore.

PR #94, opened 2016-05-15
Servercontroller

bvanwijk:
I created a serverController class, so everything that was previously stored in the main is now in the servercontroller.
Because of this change, I had to update a lot of tests and classes so they use the new servercontroller.

### DIFF
--- a/backend/src/main/java/com/benine/backend/ServerController.java
+++ b/backend/src/main/java/com/benine/backend/ServerController.java
@@ -39,7 +39,7 @@ public class ServerController {
     running = false;
     setupLogger();
 
-    cameraController = new CameraController(this);
+    cameraController = new CameraController();
     
     String user = config.getValue("sqluser");
     String password = config.getValue("sqlpassword");
@@ -133,6 +133,14 @@ public class ServerController {
    */
   public CameraController getCameraController() {
     return cameraController;
+  }
+  
+  /**
+   * Sets the cameraController.
+   * @param cameraController the cameracontroller
+   */
+  public void setCameraController(CameraController cameraController) {
+    this.cameraController = cameraController;
   }
   
   /**
--- a/backend/src/main/java/com/benine/backend/ServerController.java
+++ b/backend/src/main/java/com/benine/backend/ServerController.java
@@ -1,5 +1,7 @@
 package com.benine.backend;
 
+import java.io.File;
+
 import com.benine.backend.camera.CameraController;
 import com.benine.backend.camera.SimpleCamera;
 import com.benine.backend.database.Database;
@@ -11,6 +13,10 @@ import com.benine.backend.http.HttpController;
  * Class containing the elements to make the server work.
  */
 public class ServerController {
+  
+  private static ServerController serverController;
+  
+  private static String mainConfigPath = "configs" + File.separator + "main.conf";
   
   private Logger logger;
 
@@ -29,7 +35,7 @@ public class ServerController {
    * Sets up everything needed to run the server.
    * @param configPath path to the main config file.
    */
-  public ServerController(String configPath) {
+  private ServerController(String configPath) {
     config = setUpConfig(configPath);
     running = false;
     setupLogger();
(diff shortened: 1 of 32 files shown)


Kingdorian:
It might be a good idea to make the ServerController a singleton. Generally speaking if you want multiple servers you would just run multiple instances of the code, for a couple of reasons including security and just general process management. This would prevent all the passing of ServerController to constructors and the servercontroller attributes.

NaomidR:
Good fix that everything is now in another class than the main. I agree with merging if the test coverage for the servercontroller is good.

bvanwijk:
The line coverage for the servercontroller is 91%.
For the whole system the line coverage is now: 85%

backend/src/main/java/com/benine/backend/camera/CameraController.java
```diff
@@ -24,14 +26,10 @@
...
 
-  private ServerController serverController;
-
   /**
    * Create a new CameraController.
-   * @param serverController to interact with the server.
    */
-  public CameraController(ServerController serverController) {
-    this.serverController = serverController;
+  public CameraController() {
```
Kingdorian: It's a very tiny detail but we can remove this constructor at all since it now is the default constructor.

backend/src/main/java/com/benine/backend/camera/CameraController.java
```diff
@@ -44,6 +42,22 @@ public void addCamera(Camera camera) {
     highestIdInUse++;
     cameras.add(camera);
   }
+
```
Kingdorian: What are these two methods used for? Since I think you could just replace the call to these methods to a call to ServerController.getInstance().getCameraController() right?  but maybe I'm missing something.

backend/src/main/java/com/benine/backend/http/HttpController.java
```diff
@@ -26,27 +26,21 @@
...
-   * @param serverController the server controller that contains
-   *                      the camera's which this server interacts with.
    */
-  public HttpController(HttpServer httpserver, Logger logger, ServerController serverController) {
+  public HttpController(HttpServer httpserver, Logger logger) {
     this.logger = logger;
-    this.serverController = serverController;
     this.server = httpserver;
-
+    serverController = ServerController.getInstance();
```
Kingdorian: I think that if the serverController is a singleton we won't have to store it here as an attribute.

Kingdorian:
It looks really neat now  there is one little thing however.
There are a lot of constructors that essentially are the default constructor. Like the one that I pointed out but also the others that only make a call to super with the same parameter. We might want to remove those to achieve cleaner code.  If thats fixed I think this can be merged.

backend/src/main/java/com/benine/backend/camera/CameraController.java
bvanwijk: Yes you are right they could be replaced with the call to ServerController.getInstance(), but since they are used in this class a few times. I though it would be a cleaner to have a possibility to just call getDatabase() for example. Since they are only used in this class I can make them private methods.

Kingdorian: I was afraid things could get confusing when calling this method from other classes so making them private would be a perfect solution :)

bvanwijk:
@Kingdorian the constructors that make only a call to super with arguments can't be removed. The default constructor is a constructor without arguments, and the super class doesn't have a constructor without arguments so this is not possible.

"Note: If a constructor does not explicitly invoke a superclass constructor, the Java compiler automatically inserts a call to the no-argument constructor of the superclass. If the super class does not have a no-argument constructor, you will get a compile-time error. Object does have such a constructor, so if Object is the only superclass, there is no problem."
https://docs.oracle.com/javase/tutorial/java/IandI/super.html

Kingdorian:
@bvanwijk I'm really sorry! I forgot that was the case for the calls to the super class constructor. There is one totally empty constructor left (for the cameraController) as far as I can see.

bvanwijk:
@Kingdorian no problem, I will remove the totally empty constructor.

Kingdorian:
Great changes, I think this can be merged if the others agree.

backend/src/main/java/com/benine/backend/ServerController.java
```diff
@@ -0,0 +1,202 @@
...
+  private ServerController(String configPath) {
+    config = setUpConfig(configPath);
+    running = false;
+    setupLogger();
+
+    cameraController = new CameraController();
+    
+    String user = config.getValue("sqluser");
+    String password = config.getValue("sqlpassword");
+    database = new MySQLDatabase(user, password, logger);
```
jochem725: Maybe this could be in a method `createDatabase(String user, String pass)` to make this constructor shorter, same holds for adding the cameras.

jochem725:
Looks great! It really makes things more clear. I've added a small comment, but I agree with merging already.

PR #95, opened 2016-05-16
Preset View

jochem725:
This is a very basic version of preset add and edit views.
Actual preset adding and editing functionality is missing here, since it is another sprint task.


bvanwijk:
Looks like a good setup for the preset creating and editing view.

Kingdorian:
Looks good. Maybe we could change the color of the create preset button since in the interface green usually means something is activated  (autofocus/iris and server connection for example) this could get confusing.

jochem725:
@Kingdorian I think you're right, I've changed the edit and add button colors to blue for now.

PR #96, opened 2016-05-17
Preset keywords

Kingdorian:
I added preset keywords to the preset class.

### DIFF
--- a/backend/src/main/java/com/benine/backend/Preset.java
+++ b/backend/src/main/java/com/benine/backend/Preset.java
@@ -87,6 +87,7 @@ public class Preset {
     json.put("autoiris", autoiris);
     json.put("image", image);
     json.put("id", presetid);
+    json.put("keywords", keyWords);
 
     return json.toString();
   }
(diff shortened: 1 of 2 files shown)


bvanwijk:
Looks good, and well tested.

backend/src/main/java/com/benine/backend/Preset.java
```diff
@@ -188,23 +256,11 @@ public boolean equals(Object o) {
     if (autofocus != preset.autofocus) {
       return false;
     }
-    
+    if (!keyWords.equals(preset.getKeywords())) {
```
bvanwijk: You updated the equals method but you didn't update the hashcode method

jochem725:
Looks fine, maybe we should keep track of all keywords used, so we can provide this to the search interface later.

NaomidR:
Good that you added a list with the keywords. We'll need this later. Looks good

bvanwijk:
I think this can be merged.

PR #97, opened 2016-05-17
Cameracreator

bvanwijk:
I added a camera creator which nows which camerafactory to use for every camera type.
This camera creator uses the right camerafactories to create the corresponding camera objects.
The servercontroller reads from the mainconfig file the information about every camera and uses the cameracreator to create the specified camera.


backend/configs/main.conf
```diff
@@ -1,8 +1,9 @@
 ipadress=127001
-camera1=http://tuincam.bt.tudelft.nl/mjpg/video.mjpg
-camera2=http://131.180.123.51/zm/cgi-bin/nph-zms?mode=jpeg&monitor=2&scale=100&buffer=100
-camera3=http://131.180.123.51/zm/cgi-bin/nph-zms?mode=jpeg&monitor=1&scale=100&buffer=100
-camera2IP=123111
+
+camera_1=simplecamera, http://tuincam.bt.tudelft.nl/mjpg/video.mjpg
```
Kingdorian: Maybe it is cleaner to make this into camera_1_name and camera_1_adress and maybe even a camera_1_type.

backend/src/main/java/com/benine/backend/camera/CameraCreator.java
```diff
@@ -0,0 +1,63 @@
...
+
+  static CameraCreator cameraCreator;
+  
+  /**
+   * Map of string types to right factory.
+   */
+  static final Map<String, CameraFactory> CAMERA_TYPES;
+  
+  static {
+    CAMERA_TYPES = new HashMap<String, CameraFactory>();
```
Kingdorian: Very clean.

bvanwijk:
I changed the camera information in the config file to use multiple lines.

backend/src/main/java/com/benine/backend/camera/CameraCreator.java
```diff
@@ -0,0 +1,86 @@
...
+  /**
+  * Loads the camera's specified in the config in the camera controller.
+  */
+  public void loadCameras() {
+    Config config = ServerController.getInstance().getConfig();
+    int i = 1;
+    String type =  config.getValue("camera_" + i + "_type");   
+    while (type != null) {
+      try {
+        ServerController.getInstance().getCameraController().addCamera(createCamera(i, type));
```
jochem725: I think this is a case of method chaining, maybe you can do something about it.

backend/src/main/java/com/benine/backend/camera/CameraCreator.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param index in the config file of the camera.
+   * @param type of this camera to create.
+   * @return Camera object
+   * @throws InvalidCameraTypeException When camera object can't be created.
+   */
+  private Camera createCamera(int index, String type) throws InvalidCameraTypeException {
+    System.out.println(type);
+    CameraFactory factory = CAMERA_TYPES.get(type);
+    if (factory == null) {
+      CameraController.logger.log("The type of camera" + index + "is not specified: " + type,
```
jochem725: Maybe there should be space in `camera" + index + "is` to make sure it is printed correctly in the log.

backend/src/main/java/com/benine/backend/camera/CameraCreator.java
```diff
@@ -0,0 +1,86 @@
...
+      try {
+        ServerController.getInstance().getCameraController().addCamera(createCamera(i, type));
+      } catch (InvalidCameraTypeException e) {
+        CameraController.logger.log("Camera " + i + " in the config file can not be created",
+            LogEvent.Type.CRITICAL);
+      }
+      i++;
+      type = config.getValue("camera_" + i + "_type");
+    }  
+  }
```
NaomidR: Maybe you can split this method so that the try-catch is in another method.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCameraFactory.java
```diff
@@ -13,26 +15,21 @@
...
-  public IPCamera createCamera(String[] camSpec) throws InvalidCameraTypeException {
-    switch (camSpec[0]) {
-      case "ipcamera" : CameraController.logger.log("Create IP camera object", LogEvent.Type.INFO);
-      return createIpcamera(camSpec[1]);
-      default: CameraController.logger.log("Create IP camera object", LogEvent.Type.WARNING);
+  public IPCamera createCamera(int index) throws InvalidCameraTypeException {
+    Config config = ServerController.getInstance().getConfig();
+    String address = config.getValue("camera_" + index + "_address");
+    if ( address == null) {
+      CameraController.logger.log("Can't create Ipcamera object with specified info.",
```
jochem725: Maybe this can be more accurate by saying `This camera has no address... etc.`

Kingdorian:
The code looks very good. If I understand this correctly the cameracreator is a layer on top of the camera factories? If that's the case would CameraFactory not be a better name with subclasses for Ipcamera and such? Or could it just be incorporated in the CameraController class?

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
```diff
@@ -28,11 +27,6 @@ public CameraInfoHandler(Logger logger) {
   public void handle(HttpExchange exchange) throws IOException {
     getLogger().log("Got an http request with uri: "
             + exchange.getRequestURI(), LogEvent.Type.INFO);
-    try {
-      respond(exchange, getCameraController().getCamerasJSON());
-    } catch (CameraConnectionException e) {
-      getLogger().log("Cannot connect to camera with id: " + e.getCamId(), LogEvent.Type.CRITICAL);
-    }
-    respond(exchange, "{\"succes\":false}");
+    respond(exchange, getCameraController().getCamerasJSON());
```
Kingdorian: I might be wrong but since getCamerasJSON does not return succes false when it fails to connect the camera it would be unclear to the client if the camera is not existent or if its unreachable.

bvanwijk:
@Kingdorian Yes, the cameracreator  decides which factory to use. It could be incoporated in the cameraController class, but I think using a seperate class for it is better, Because the cameracontroller now does not have to know about the different types of cameras, and it will reduce the responsibilities of the cameracontroller. CameraFactory is the abstract class the different concrete camerafactories such as IPCameraFactory implement.

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
bvanwijk: That is true but when there is only one camera unreachable, I think the others should still be usable. So previously when the getCamerasJSON fails it means at least one camera fails, but the information of the other cameras should still be sent to the client.

Kingdorian:
Then we already have the abstract factory design pattern but not fully. I recommend implementing the full abstract factory pattern with a factoryproducer that is responsible for returning a camerafactory which then in is used in the cameracontroller.
http://www.tutorialspoint.com/design_pattern/images/abstractfactory_pattern_uml_diagram.jpg

bvanwijk:
@Kingdorian So if I understand you correctly, you mean the cameracontroller has to ask the CameraFactoryProducer with a string type from the config file for the right cameraFactory for every camera.

Kingdorian:
Yes, partly. The cameraCreator would change to a CameraFactoryProducar that produces a CameraFactory for each camera and then the CameraController uses this to create a camera.

bvanwijk:
@Kingdorian This makes it better I think, I changed the Cameracreator to a cameraFactoryProducer.

Kingdorian:
Looks great! Just remember to use git mv instead of just renaming a file, that way the history of the file is preserved :)

jochem725:
Looks fine! I think this can be merged.

PR #99, opened 2016-05-17
Fix missing call to connect to the database server.

bvanwijk:


egedebruin:
Well noticed, I agree with merging

NaomidR:
This looks good, this can be merged.

jochem725:
Great change, I think this can be merged.

Issue #83, opened 2016-05-13
Make Logger.log(Exception e) method in Logger


Kingdorian:

bvanwijk:
Fixed in #100

PR #100, opened 2016-05-18
Logger exception support

Kingdorian:
As mentioned in https://github.com/Kingdorian/Contextproject-BeNine/issues/83 there is not much to tell about this really.


jochem725:

bvanwijk:

PR #101, opened 2016-05-18
Response message

### LINKED ISSUE #71: Use responseSuccess/Failure everywhere
This is a new method and it is probably not used everywhere yet. This has to be checked.

NaomidR:
Changed all classes so that it now uses the response message. The test coverage for the http package is now 91%. This fixes #71


backend/src/main/java/com/benine/backend/http/RecallPresetHandler.java
```diff
@@ -57,13 +56,17 @@ public void handle(HttpExchange exchange) throws IOException {
...
-  public void movingCamera(IPCamera ipcamera, Preset preset) throws CameraConnectionException {
-    Position position = preset.getPosition();
-    ipcamera.moveTo(position, preset.getPanspeed(), preset.getTiltspeed());
-    ipcamera.zoomTo(preset.getZoom());
-    ipcamera.moveFocus(preset.getFocus());
-    ipcamera.setAutoFocusOn(preset.isAutofocus());
-    ipcamera.setIrisPos(preset.getIris());
-    ipcamera.setAutoIrisOn(preset.isAutoiris());
+  public void movingCamera(IPCamera ipcamera, Preset preset, HttpExchange exchange) {
+    try {
```
bvanwijk: Why do you do the try catch in this method, I think it would be cleaner to just throw the CameraConnectionException and catch it in the handle method. You can then do the ResponeFailure(exchange) in that method.

Kingdorian:
I think it would be cleaner to call the method respondFailure instead of responseFailure since it implies an action (method) and not a thing "response".

NaomidR:
I think that it is a response to the user about what happened, a success or a failure, so that's why I named it response. I don't really think that it matters that much if it is called response or respond, for both names arguments can be thought of. But as for you really want this changed, I will. Next time please mention this earlier so that I can fix it everywhere correct the first time. The method 'responseFailure' and 'responseSuccess' already existed in the previous sprint if I'm not mistaken. If not then I take back my words ;)

Kingdorian:
You're right I should've said this earlier. However its just a simple IDE refactor so this should be able to do in a couple of clicks +  a couple of minutes for chaning javadoc in some places.

backend/src/main/java/com/benine/backend/http/RecallPresetHandler.java
NaomidR: I think I moved the try catch statement to the movingCamera method to make the handle method shorter and to avoid having a lot of catch statements there. Do you wanna see something like try { movingCamera(ipcamera,preset) } catch { responseFailure(exchange)} in the handle method? That would be shorter indeed.

backend/src/main/java/com/benine/backend/http/RecallPresetHandler.java
```diff
@@ -57,13 +56,17 @@ public void handle(HttpExchange exchange) throws IOException {
...
+      Position position = preset.getPosition();
+      ipcamera.moveTo(position, preset.getPanspeed(), preset.getTiltspeed());
+      ipcamera.zoomTo(preset.getZoom());
+      ipcamera.moveFocus(preset.getFocus());
+      ipcamera.setAutoFocusOn(preset.isAutofocus());
+      ipcamera.setIrisPos(preset.getIris());
+      ipcamera.setAutoIrisOn(preset.isAutoiris());
+    } catch (CameraConnectionException e) {
+      responseFailure(exchange);
```
Kingdorian: Could you please log here that the exception occurred?

backend/src/main/java/com/benine/backend/http/RecallPresetHandler.java
bvanwijk: Yes, I think that is a better solution.
Since you already have a try catch in the handle method you can even do it like: 
try { 
....
  movingCamera(ipcamera, preset)
} catch( SQLException | CameraConnectionException e ) { 
  ...
}

NaomidR: Oh that's right, that would be better indeed. I'll fix it, thanks!

backend/src/main/java/com/benine/backend/http/RecallPresetHandler.java
NaomidR: Fixed this

NaomidR:
@Kingdorian yes of course it will be just a few clicks, just wanted to point it out to you for the next time. I'll of course change this, no problem at all.

Kingdorian:
I'll try to point it out earlier next time :) I just didn't notice it earlier.

jochem725:
Seems fine, I agree with @Kingdorian on the response method names. `respondFalse` and `respondTrue` would make it more clear that the handler is the entity responding, this can easily be fixed.

bvanwijk:
I agree with merging, looks good.

PR #102, opened 2016-05-18
Fix checkstyle and findbugs

NaomidR:
The few checkstyle and PMD warnings in develop are now  gone.

### DIFF
--- a/backend/src/main/java/com/benine/backend/Preset.java
+++ b/backend/src/main/java/com/benine/backend/Preset.java
@@ -6,7 +6,6 @@ import org.json.simple.JSONObject;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.StringJoiner;
 
 /**
  * A moving preset to be able to add to the database.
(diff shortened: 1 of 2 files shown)


Kingdorian:
I think these changes are a bit to small for a complete pullrequest since its just a  tiny refactor this can be done when making changes to a class when working on a feature. We can merge this but in the future I  think we should avoid this kind of "fix checkstyle" or "small refactor" pullrequests and just icorporate the changes in larger pull requests. This avoids clutter of unused pull requests like this one.

NaomidR:
You commented this in the other pull request as well and I can understand why you're saying this. However, I discussed this with @jochem725 this morning and we decided to just make a pull request so that these errors are fixed. There's no need to go over this in very much depth though.

Kingdorian:
For the other pull request I made this comment for I now understand why that should be a pullrequest.

Just discussed this with @jochem725  too and he indeed told me you discussed this but he agreed with me on this being to small of a change for a pullrequest.

bvanwijk:
I agree with merging.

Issue #86, opened 2016-05-13
Slider acting as a JoyStick


jochem725:
The sliders should act like a JoyStick to make more precise adjustments possible.

Kingdorian:
#103

PR #103, opened 2016-05-18
Sliders acting as 1d joystick

Kingdorian:
As mentioned in https://github.com/Kingdorian/Contextproject-BeNine/issues/86. Currently the focus and iris sliders only zoom in or out, not with a certain speed. This will be fixed when fixing the maximum amount of api  calls and commands send to the camera are limited.

### DIFF
--- a/backend/src/main/java/com/benine/backend/http/FocussingHandler.java
+++ b/backend/src/main/java/com/benine/backend/http/FocussingHandler.java
@@ -54,7 +54,7 @@ public class FocussingHandler extends RequestHandler {
       if (setPos != null) {
         focusCam.setFocusPos(Integer.parseInt(setPos));
       } if(speed != null) {
-        focusCam.moveFocus((int)(50 + (9.9*Integer.parseInt(speed))));
+        focusCam.moveFocus((int)(50 + (4.45*Integer.parseInt(speed))));
       }
         response = "{\"succes\":\"true\"}";
     } catch (CameraConnectionException e) {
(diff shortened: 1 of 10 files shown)


backend/src/main/java/com/benine/backend/http/FocussingHandler.java
```diff
@@ -45,15 +45,18 @@ public void handle(HttpExchange exchange) throws IOException {
...
+    String speed = parsedURI.getValue("speed");
     try {
       if (autoOn != null) {
         boolean autoOnBool = Boolean.parseBoolean(autoOn);
         focusCam.setAutoFocusOn(autoOnBool);
       }
       if (setPos != null) {
         focusCam.setFocusPos(Integer.parseInt(setPos));
+      } if(speed != null) {
+        focusCam.moveFocus((int)(50 + (4.45*Integer.parseInt(speed))));
```
bvanwijk: Where does the 4.45 come from?

Kingdorian: To map the slider domain (which is -10 to 10) to a domain of 0-99 which is 4.95 larger so this should actually be 4.95 I'll fix it.

bvanwijk: Yes, I think it would be better to do it client side, because otherwise it could get confusing in the future when you change one thing and it's maped a few times to another domain.

Kingdorian:
@bvanwijk Fixed.

bvanwijk:
Nice, I agree with merging.

jochem725:
Seem fine, I think this can be merged.

egedebruin:
Looks good, i agree with merging

PR #104, opened 2016-05-18
Limit slider commands

bvanwijk:
The focus, zoom and iris inputs can now only send commands to the backend once in 130 ms.


jochem725:
Seems fine however I think we need some security for this on the backend server to guarantee a code change in the future will not flood a camera.

bvanwijk:
@jochem725 I think that is part of another task on the sprint of this week.

frontend/public/js/control.js
```diff
@@ -158,36 +159,91 @@ function sendMove(){
...
 	$('#auto_focus').removeClass("btn-success");
-	$.get("/api/backend/camera/" + currentcamera + "/focus?autoFocusOn=false&position=" + focus , function(data) {});
-	console.log("Focus: " + focus);
+	if (focusSend === false) {
+		focus = f;
+		focusSend = true;
+		setTimeout(function(){sendFocus(); focusSend = false;}, 130);
+		sendFocus();
+	} else {
+		focus = (parseInt(f) + parseInt(focus)) / 2;
```
Kingdorian: I see this code (strucure) 3 times in the code. Maybe you can turn this in to a more generalized method to get rid of the duplicate code.

bvanwijk: @Kingdorian Yes, I agree this is duplicate code. I changed it, do you agree with merging now?

Kingdorian:
Great I think this can be merged now!

PR #105, opened 2016-05-18
MJPEG Stream Reader

jochem725:
We can now read Motion JPEG streams into the server.
The server keeps track of the streams, by first looking for a MJPEG header, then the JPEG belonging to that header, storing the last image so it can be retrieved.

The `CameraController` now contains a `StreamController`, from which you can get a `MJPEGStreamReader` which contains a method to take a snapshot of the stream.


backend/src/main/java/com/benine/backend/video/Stream.java
```diff
@@ -0,0 +1,57 @@
...
+public class Stream {
+
+  private URLConnection connection;
+  private InputStream inputstream;
+
+  /**
+   * Constructor for a new stream object.
+   * @param streamurl The url to get the stream from.
+   * @throws IOException If an exception occurs while creating the stream,
+   * rendering the stream useless.
```
bvanwijk: This gives a small checkstyle warning, rendering should be right under IOException.

bvanwijk:
Looks good, nice to see no findbugs and pmd warnings.

backend/src/main/java/com/benine/backend/video/Stream.java
jochem725: Nice find @bvanwijk, I've fixed it.

backend/src/main/java/com/benine/backend/video/MJPEGStreamReader.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param amount The amount of bytes to check.
+   * @return An int[] array containing the checked bytes in order.
+   * @throws IOException when the next bytes cannot be read from the stream.
+   */
+  private int[] checkNextBytes(int amount) throws IOException {
+    int[] byteArray = new int[amount];
+
+    bufferedStream.mark(amount);
+    for (int i = 0; i < amount; i++) {
+      byteArray[i] = bufferedStream.read();
```
Kingdorian: Wouldn't it be easier to use the  read(byte[] b, int off, int len) method and read all the bytes at once? 
https://docs.oracle.com/javase/7/docs/api/java/io/BufferedInputStream.html#read(byte[],%20int,%20int)

backend/src/main/java/com/benine/backend/video/StreamController.java
```diff
@@ -0,0 +1,125 @@
...
+   * @param streams Map for integer, thread
+   * @param threads Map for integer, thread
+   */
+  public StreamController(Map<Integer, StreamReader> streams, Map<Integer, Thread> threads) {
+    this.streams = streams;
+    this.threads = threads;
+  }
+
+  /**
+   * Adds a camera to the HashMap containing all streams if there is a stream available.
```
Kingdorian: I think HashMap should be Map.

backend/src/main/java/com/benine/backend/video/StreamController.java
```diff
@@ -0,0 +1,125 @@
...
+
+    return streamLink;
+  }
+
+  /**
+   * Gets the type and sets an ENUM if it matches.
+   * @param   cam The camera from which the stream is fetched.
+   * @return  A StreamType Enum.
+   */
+  private StreamType verifyType(Camera cam) {
```
Kingdorian: Wouldn't it be easier to store a streamType attribute in the camera class?

backend/src/main/java/com/benine/backend/video/StreamNotAvailableException.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Defines an exception thrown when a stream is not available.
+ */
+public class StreamNotAvailableException extends Exception {
+
+  /**
+   * Should be thrown if a stream is not available.
+   */
+  public StreamNotAvailableException() {}
```
Kingdorian: Maybe you could provide some info about the Stream like, the stream for what camera or what link and maybe even a reason why its not available for example.

backend/src/main/java/com/benine/backend/video/MJPEGStreamReader.java
jochem725: I've been thinking about that, but reading from an input stream consumes the input, so looking for the header by reading consumes the first 2 bytes of the JPEG image (the flag). I think it is more clear to "peek" into the next 2 bytes instead of reading them, because if they are read they need to be re-added in front of the JPEG image.

backend/src/main/java/com/benine/backend/video/StreamController.java
jochem725: I think that would indeed be a much better idea, I'll implement it.

backend/src/main/java/com/benine/backend/video/MJPEGStreamReader.java
Kingdorian: Okay then this is indeed a better solution.

jochem725:
I've refactored Camera to improve code reuse and added the stream type to Camera.

backend/src/main/java/com/benine/backend/video/StreamNotAvailableException.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Defines an exception thrown when a stream is not available.
+ */
+public class StreamNotAvailableException extends Exception {
+
+  private int camid;
+
+  /**
+   * Should be thrown if a stream is not available.
```
bvanwijk: You are missing the camid parameter in the documentation

bvanwijk:
Looks like a good refactor for the camera classes.

backend/src/main/java/com/benine/backend/video/StreamNotAvailableException.java
```diff
@@ -5,9 +5,20 @@
  */
 public class StreamNotAvailableException extends Exception {
```
Kingdorian: It would be usefull to add a toString method.

jochem725:
@Kingdorian Great idea, I've added it. @bvanwijk I've updated the JavaDoc.

Kingdorian:
I think this can be merged.

PR #106, opened 2016-05-18
Load from database

egedebruin:
I have added that the backend will now load all de presets and cameras from the database on startup.
The only thing which I do not really know how to fix is that only IPCameras can be added and gotten from the database, since there is an IPaddress column in the database. So for now only IPcameras are added to the database.


backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -266,6 +268,38 @@ public void addCamera(int id, String ip) {
   }
 
   @Override
+  public ArrayList<Camera> getAllCameras() throws SQLException {
```
jochem725: Maybe this method is a bit long. Is it possible to split it into smaller methods?

backend/src/main/java/com/benine/backend/ServerController.java
```diff
@@ -64,9 +65,14 @@ public static synchronized ServerController getInstance() {
    */
   public void start() {
     httpController = new HttpController(config.getValue("serverip"),
-        Integer.parseInt(config.getValue("serverport")), logger); 
-    
+        Integer.parseInt(config.getValue("serverport")), logger);
     startupDatabase();
+    try {
```
jochem725: If this try block fails, the application becomes useless because there is no data. Maybe we should throw an exception here to notify the main about a start() failure.

jochem725:
I've added some comments. I think the test coverage of the MySQLDatabase class is too low for this to be merged into develop. Also there is a small PMD bug which can be fixed easy.

NaomidR:
I agree with @jochem725 about e.g. the method being too long. I also commented something about the test coverage in another branch. There is also one findbugs warning about the ResultSet not being cleaned up. Can you have a look at this if it can be fixed?

egedebruin:
I have changed some things where I needed to merge from other branches, which may lead to some unclarity.

Now if the program is startup, the program checks if the cameras in the database correspond to the cameras in the program. If not, the database is changed. So the cameras in de database are always on track to the cameras in the program.

On the testing in the MySQLDatabase, I did not find a way the test resultsets, as they are only made when querying to a correct database (not a mocked one), so that causes a lot of non testable code.

egedebruin:
This will be merged in another branch.

egedebruin:
This will still be merged in develop.

PR #108, opened 2016-05-19
Image preset

NaomidR:
An image is now created, rescaled and saved when a preset is made.  Also, the path of the image is sent to the preset. I added the images to the gitignore.


backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -40,20 +48,43 @@ public void handle(HttpExchange exchange) throws IOException {
...
+   */
+  public static void createImage(Preset preset, int cameraID, int presetID) throws StreamNotAvailableException, IOException {
+    ServerController serverController = ServerController.getInstance();
+    StreamReader streamReader = serverController.getStreamController().getStreamReader(cameraID);
+    BufferedImage bufferedImage = streamReader.getSnapShot(); 
+    
+    //Rescale image so it loads faster.
+    //BufferedImage buffer = (BufferedImage)bufferedImage.getScaledInstance(360, 235, BufferedImage.SCALE_DEFAULT);
+    
+    File path = new File("static" + File.separator + "presets" + File.separator + cameraID + "_" + presetID + ".jpg");
```
bvanwijk: I think the path should be File.separator + "static" + File.separator + "presetImages" + File.separator + cameraID + "_" + presetID + ".jpg"

NaomidR: I saw that too, it is already fixed! At least, it is now going to the presets file, which already existed.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -40,20 +48,43 @@ public void handle(HttpExchange exchange) throws IOException {
...
+        
+        //Create corresponding image
+        createImage(preset, cameraID, presetID);
         responseSuccess(exchange);
       
       }
-    } catch (SQLException e) {
-      System.out.println(e.toString());
+    } catch (SQLException | StreamNotAvailableException e) {
       getLogger().log("Preset can not be added to the database", LogEvent.Type.CRITICAL);
```
jochem725: Maybe a reason why the preset cannot be added to the database should be added here.

jochem725:
Seems fine, but there are some Checkstyle and PMD warnings which can be resolved. I also added a comment regarding logging a thrown exception.

backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
NaomidR: So you mean splitting the two exceptions and saying that or the stream wasn't availble or there was a sql exception?

jochem725: Yes, something like that.
Because reading the current message in the log is not very informative.

Kingdorian:
Looks good but we don't we merge this in Jochem's branch?

NaomidR:
@Kingdorian  hmm that's possible as well yes. I did merge Jochems branch into mine but if he made changes after that then it would be a good idea indeed.

NaomidR:
@jochem725 I fixed the warnings.

jochem725:
@Kingdorian @NaomidR The branch you are referring to has already been merged into Develop in #105.

Kingdorian:
Ah so its just github displaying the wrong commits to merge then?

jochem725:
@Kingdorian Yes, GitHub does not keep track of already merged commits. You can view the changes by comparing though, https://github.com/Kingdorian/Contextproject-BeNine/compare/develop...imagePreset

bvanwijk:
I agree with merging.

Kingdorian:

PR #109, opened 2016-05-19
Change database

egedebruin:
I have changed the database, now there are only two tables, the camera table and presets table.


backend/src/main/java/com/benine/backend/database/Database.java
```diff
@@ -22,40 +22,25 @@
...
+  void addPreset(Preset preset) throws SQLException;
 
   /**
    * Delete a preset from the database.
    * @param camera ID of the camera
-   * @param cameraPresetNumber ID of the preset for the camera
    * @throws SQLException No right connection found
    */
-  void deletePreset(int camera, int cameraPresetNumber) throws SQLException;
+  void deletePreset(int camera) throws SQLException;
```
Kingdorian: I think camera should be presetID?

NaomidR:
The test coverage for the MySQLDatabase is still 62%, I think that this should become higher (in the future, if possible).

backend/src/main/java/com/benine/backend/database/Database.java
NaomidR: That would indeed make more sense as for a camera can have multiple presets so now it is not clear which preset you are deleting?

egedebruin: Well noticed, I looked over that.

egedebruin:
The coverage will be slightly increased in another branch, but it is not possible to get it very high. The database can not be fully unit tested, as a lot of the database can only be tested if there is a real database. In testing we use a mock database, which only verifies the right queries are handled.

jochem725:
@egedebruin Which branch exactly increases the coverage of the database? Because I can still see some cases which can be tested easily.

egedebruin:
I am busy with it in the loadFromDatabase branch.

bvanwijk:
I agree with merging, if you increase the coverage in the other branch you mentoined

Kingdorian:
I think we can merge this if the coverage is increased by another branch.

PR #110, opened 2016-05-19
Single logger

bvanwijk:
Since the Servercontroller is singleton when can get the logger from there, so I removed the passing of the logger instance to the different classes.


Kingdorian:
There is nothing to say about this really, except it really cleans up the code.

egedebruin:
Nice code fix, it is a lot cleaner.

jochem725:
Great! This really makes the logger easier to use.

PR #111, opened 2016-05-20
Fix ipcamera command format

bvanwijk:
This fixes the formatting of the IPCamera commands.
I have also remove the author notation in every class.


NaomidR:
Good processing of the feedback. The changes look good, also no warnings so I agree with merging.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera.java
```diff
@@ -402,13 +405,13 @@ public String toJSON() throws CameraConnectionException {
...
     json.put("id", this.getId());
     try {
-      json.put("pan", getPosition().getPan());
-      json.put("tilt", getPosition().getTilt());
-      json.put("zoom", (double) getZoomPosition());
-      json.put("focus", (double) getFocusPosition());
-      json.put("autofocus", isAutoFocusOn());
-      json.put("iris", (double) getIrisPosition());
-      json.put("autoiris", isAutoIrisOn());
+      json.put("pan", new Double(getPosition().getPan()));
```
jochem725: Why is it required to create a new object from an already existing double object?

bvanwijk: I've no idea, I think I choose the wrong version due to a merge conflict.

jochem725:
I think this can be merged.

PR #112, opened 2016-05-20
MJPEG Stream Distributer

jochem725:
I've implemented a MJPEG stream distribution endpoint.
`localhost:3725/camera/{id}/mjpeg` should return an image stream.

I don't think we should merge this branch for this iteration though.
Due to the blocking nature of the current server, I had to implement a different Server called Jetty, which works Asynchronously. I propose to add a refactor of the server to Jetty to the next sprint, so the complete API becomes asynchronous and we only have a single server running. Due to this proposal, I have not yet supplied tests.

Please let me know if the structure is correct, and if you agree with the refactoring proposal.


backend/pom.xml
```diff
@@ -1,182 +1,187 @@
...
-            <version>${findbugs.plugin.version}</version>
-         </plugin>
-         <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-pmd-plugin</artifactId>
-            <version>${pmd.plugin.version}</version>
-         </plugin>
-      </plugins>
-   </reporting>
+    <modelVersion>4.0.0</modelVersion>
```
jochem725: I don't think it was a good idea to commit a reformat of the pom.xml file in this branch.

bvanwijk:
Looks good, I agree with a refactor of the server.
Just remember to remove your name.

backend/src/main/java/com/benine/backend/http/jetty/CameraRequestHandler.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+  /**
+   * Fetches camera id from http exchange.
+   * @param request the request to fix the id from.
+   * @return the id of the camera.
+   */
+  public int getCameraId(Request request) {
+    // Get path
+    Pattern pattern = Pattern.compile(".*/(\\d*)/.*");
```
Kingdorian: I think this might go wrong when when an request is made to /camera/1/preset/3/. The pattern .*/camera/(\\d*)/.* or something similar might be better suited.

Kingdorian:
Looks good. But I think the events that happen should be logged a bit more extensive. There is barely no logging at all at the moment.

jochem725:
@Kingdorian I've opened a #123 for the http-refactor. Maybe we should first handle that pull request and then return to this one, since some major changes have occurred.

jochem725:
I've just merged the http-refactor into this branch.
Please verify everything is working.

@bvanwijk It appears after the merge, before every API call commands are sent to the camera. I'm not sure where this originates from, could you provide some insight on this?

jochem725:
I've solved the issue stated in my previous comment. If toJSON() cannot connect to a camera it blocks, we might want to do this smarter in a future iteration.

Kingdorian:
Have you verified this handles simultaneous HTTP requests in a safe way? So it does not interact with the data with with multiple threads at the same time?

jochem725:
@Kingdorian The asynchronous part of the server should only apply to receiving and sending data not the interaction with the data itself. I have not been able to identify cases in which multiple interactions with the data take place at the same time.

Kingdorian:
Alright that's very nice then! Do you have any documentation on (which we can make a reference to in the architecture deisgn document) that or does it rely on your testing?

bvanwijk:
Looks very good, also a very good test coverage for the refactored http handlers

egedebruin:
Very good work, I also think this can be merged.

jochem725:
@Kingdorian Currently it is based on my testing, and some documentation about NIO. I'm still looking for a solid reference to support this in the architecture.

Kingdorian:
I took a look in the Jetty source code. Where I encountered this:

> /* ------------------------------------------------------------ */
/** Jetty HTTP Servlet Server.
 * This class is the main class for the Jetty HTTP Servlet server.
 * It aggregates Connectors (HTTP request receivers) and request Handlers.
 * The server is itself a handler and a ThreadPool.  Connectors use the ThreadPool methods
 * to run jobs that will eventually call the handle method.
 */

Which indicates the handle methods are called asynchronously I'm afraid but I'm not a 100% sure.

https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java

jochem725:
@Kingdorian You are right, after additional testing I discovered the handlers are not thread safe. I've opened #130 so we can address this in the future.

Kingdorian:
For now I think it would be already a bit better to have a volatile attribute/getter for the servercontroller in the requesthandler class.

jochem725:
@Kingdorian I've changed the ServerController singleton pattern to use volatile, I've also updated the issue with the possible synchronization solution we spoke about.

Kingdorian:
Alright it isn't an enormous problem at the moment so I think we can leave it for next sprint. I think we can merge this then.

PR #113, opened 2016-05-20
Load from database

egedebruin:
I have changed some things where I needed to merge from other branches, which may lead to some unclarity.

Now if the program is startup, the program checks if the cameras in the database correspond to the cameras in the program. If not, the database is changed. So the cameras in de database are always on track to the cameras in the program.

On the testing in the MySQLDatabase, I did not find a way the test resultsets, as they are only made when querying to a correct database (not a mocked one), so that causes a lot of non testable code.


Kingdorian:
Maybe you should merge develop into changeDatabase first.

PR #114, opened 2016-05-20
Get preset by key word

Kingdorian:


bvanwijk:
Seems fine, I agree with merging

egedebruin:
There are some PMD warnings which can be easily fixed, but other than that I agree with merging.

jochem725:

PR #115, opened 2016-05-20
Update architecture design

bvanwijk:
Update for the architecture design from share latex


egedebruin:

jochem725:
I've updated the file, since it was not the latest from ShareLatex. I agree with merging now.

PR #116, opened 2016-05-20
Preset control ui

bvanwijk:
The calls to create a preset and searching for presets are implemented.
Adding keywords when creating a preset is not yet possible.


backend/src/main/java/com/benine/backend/http/PresetCreationHandler.java
```diff
@@ -22,6 +22,8 @@
      * @throws IOException when an error occurs with responding to the request.
   */
   public void handle(HttpExchange exchange) throws IOException {
+    getLogger().log("Got an http request with the following uri in Prest creation: "
```
jochem725: There is a typo in this log message.

backend/src/main/java/com/benine/backend/http/PresetHandler.java
```diff
@@ -24,30 +26,33 @@ public void handle(HttpExchange exchange) throws IOException {
...
       }
+
       
       JSONObject jsonObject = new JSONObject();
       jsonObject.put("presets", json);
       response = jsonObject.toString();
-      
-    } catch (SQLException e) {
+    } catch (MalformedURIException | SQLException e) {
       getLogger().log("Exception occured while respoinding to the request with URI: "
```
jochem725: There is also an error in this log message.

frontend/public/css/style.css
```diff
@@ -225,6 +234,13 @@ input[type=range]:focus {
 }
 
 /* Preset area configuration. */
+#tagsearch {
+	width: 100%;
```
jochem725: Please watch the indentation of the CSS files as they may become messy if not properly indented.

jochem725:
I've added some comments, overall I agree with merging.

PR #117, opened 2016-05-20
Reduce test filesize to a minimum

jochem725:
I've reduces the MJPEG test file sizes, to hopefully make Travis no longer crash.


bvanwijk:
I agree with merging.

Kingdorian:
There is not much to say about this.

PR #121, opened 2016-05-20
Fix the currently not visible camera's and working UI

bvanwijk:
Fix for displaying the client view and creating a preset.


Kingdorian:
Looks good.

egedebruin:
Good fix.

jochem725:
This works again! Thanks for fixing it so quickly @bvanwijk.

PR #122, opened 2016-05-22
Verify ipcamera interaction

bvanwijk:
I have verified all the different calls from the backend to the camera from polycast are working.
Where necessary I have changed the calls. 
The response of every request to the camera is now also verified.


backend/src/main/java/com/benine/backend/camera/CameraController.java
```diff
@@ -76,18 +72,11 @@ private StreamController getStreamController() {
   }
 
   /**
-   * Sets up the logger.
+   * Returns the single instance of the logger.
```
jochem725: Maybe this should be singleton instead of single

jochem725:
Seems fine, I've added a small comment regarding a comment, but I think this can be merged.

NaomidR:
Looks good, great that all the calls are working now. I agree with merging this.

backend/src/main/java/com/benine/backend/camera/BasicCamera.java
```diff
@@ -1,12 +1,7 @@
...
-
-import com.benine.backend.Preset;
 import com.benine.backend.video.StreamType;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-
 /**
  * Interface for communication with remote camera's.
  * @author Bryan
```
egedebruin: We are working as a team so you should replace Bryan with Benine or remove your name.

egedebruin:
Looks very good, I agree with merging this.

Kingdorian:
Looks good, I think we can merge this but I think it would be good to wait for @jochem725's pulrequest to  be merged into develop (merge conflict wise).

backend/src/main/java/com/benine/backend/camera/BasicCamera.java
bvanwijk: Well noticed I will remove it.

PR #124, opened 2016-05-23
Http Refactor

jochem725:
This branch provides the HTTP Refactor. This should be merged in mjpeg-stream-distribute, from which it origins.

Please verify carefully if the API is still acting the same.
Everything now first passes a LogHandler, which logs the URI. In the future maybe this handler can also do the URI verification.


backend/.checkstyle
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <fileset-config file-format-version="1.2.0" simple-config="false" sync-formatter="false">
-  <local-check-config name="maven-checkstyle-plugin verify-style" location="file:/D:/Git/Contextproject-BeNine/checkstyle.xml" type="remote" description="maven-checkstyle-plugin configuration verify-style">
+  <local-check-config name="maven-checkstyle-plugin verify-style" location="file:/C:/Users/bryan/Documents/Git/Contextproject-BeNine/checkstyle.xml" type="remote" description="maven-checkstyle-plugin configuration verify-style">
```
Kingdorian: Shouldn't this be a more generic name instead of the path that obviously only works on Bryan's machine?

backend/src/main/java/com/benine/backend/http/CameraFocusHandler.java
```diff
@@ -0,0 +1,50 @@
+package com.benine.backend.http;
+
+import com.benine.backend.LogEvent;
+import com.benine.backend.camera.CameraConnectionException;
+import com.benine.backend.camera.FocussingCamera;
+import org.eclipse.jetty.server.Request;
+
+import java.io.IOException;
```
Kingdorian: I see jetty and javafx imports for http things should these two be mixed together?

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
```diff
@@ -1,24 +1,124 @@
...
+  private CameraMovingHandler moveHandler;
+
+  private CameraIrisHandler irisHandler;
+
+  private CameraZoomHandler zoomHandler;
+
+  /**
+   * Constructor for a new CameraInfoHandler, handling the /camera/ request.
+   */
+  public CameraInfoHandler() {
```
Kingdorian: Wouldn't it be better to just have a list of handlers?

backend/src/main/java/com/benine/backend/http/CameraInfoHandler.java
```diff
@@ -1,24 +1,124 @@
...
+
+    String cameraInfo = getCameraController().getCamerasJSON();
+
+    boolean routed = false;
+
+    if (checkValidCameraID(request)) {
+      String route = getRoute(request);
+      Camera cam = getCameraController().getCameraById(getCameraId(request));
+
+      switch (route) {
```
Kingdorian: I see lots of duplicate code here. Wouldn't it be better to store a string in each handler, create a local var handler and then loop over the list of handlers setting the local var handler to the handler with the string matching to route and then handle the request with that handler.

backend/src/main/java/com/benine/backend/http/CameraRequestHandler.java
```diff
@@ -0,0 +1,42 @@
...
+public abstract class CameraRequestHandler extends RequestHandler {
+
+  /**
+   * Fetches camera id from http exchange.
+   * @param request the request to fix the id from.
+   * @return the id of the camera.
+   */
+  public int getCameraId(Request request) {
+    // Get path
+    Pattern pattern = Pattern.compile(".*/(\\d*)/.*");
```
Kingdorian: I think this also accepts wrong strings like /preset/1/camera/4/, I would change this to something like .*/camera/(\\d*)//.

backend/src/main/java/com/benine/backend/http/CameraRequestHandler.java
```diff
@@ -0,0 +1,42 @@
...
+    String path = request.getPathInfo();
+
+    Matcher m = pattern.matcher(path);
+
+    if (m.matches()) {
```
Kingdorian: You can clean this up by doing something like "return m.matches ? Intteger.parseInt(m.group(1) : -1"

backend/.checkstyle
bvanwijk: This file is removed from git in https://github.com/Kingdorian/Contextproject-BeNine/pull/122/commits/956ca842581d7ae6b751f9cd84299bae10317601

backend/src/main/java/com/benine/backend/http/CameraStreamHandler.java
```diff
@@ -0,0 +1,88 @@
...
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
+          throws IOException, ServletException {
+
+    int camID = getCameraId(request);
+
+    StreamReader streamReader = null;
+    try {
+      streamReader = ServerController.getInstance().getStreamController().getStreamReader(camID);
+    } catch (StreamNotAvailableException e) {
+      e.printStackTrace();
```
Kingdorian: Please log an exception here like "getLogger.log([String:description], e);"  this also automatically logs the stacktrace (and is thus stored in the logs).

backend/src/main/java/com/benine/backend/http/CameraStreamHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    request.setHandled(true);
+  }
+
+  /**
+   * Sets the HTTP Headers so the browser detects MJPEG.
+   * @param reader                The MJPEG stream reader containing the boundary.
+   * @param httpServletResponse   The response for which the headers should be set.
+   */
+  private void setHeaders(MJPEGStreamReader reader, HttpServletResponse httpServletResponse) {
+    httpServletResponse.setContentType("multipart/x-mixed-replace;boundary="
```
Kingdorian: I don't really like the idea of hardcoding this. Maybe you can make a simple file like with the default headers for example which then can be read by a simple BufferedReader.

backend/src/main/java/com/benine/backend/http/CameraZoomHandler.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
+          throws IOException, ServletException {
+    int camID = getCameraId(request);
+
+    ZoomingCamera zoomingCam = (ZoomingCamera) getCameraController().getCameraById(camID);
+    String zoomType = request.getParameter("zoomType");
+    String zoom = request.getParameter("zoom");
+
+    System.out.println(zoomType);
```
Kingdorian: I think we've got a logger for this.

backend/src/main/java/com/benine/backend/http/PresetsCreatePresetHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        throw new MalformedURIException("No Camera ID Specified.");
+      }
+
+      CameraController cameraController = ServerController.getInstance().getCameraController();
+      Camera camera = cameraController.getCameraById(Integer.parseInt(camID));
+      System.out.println(ServerController.getInstance().getCameraController().getCameraById(1));
+
+      if (camera instanceof IPCamera) {
+        IPCamera ipcam = (IPCamera) camera;
+        System.out.println(ipcam);
```
Kingdorian: System.out again? I think it would be better to either remove this or change this to the logger.

backend/src/main/java/com/benine/backend/http/PresetsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+
+  private PresetsCreatePresetHandler createPreset;
+
+  private PresetsRecallPresetHandler recallPreset;
+
+  /**
+   * Constructor for a new PresetsHandler, handling the /presets/ request.
+   */
+  public PresetsHandler() {
+    this.createPreset = new PresetsCreatePresetHandler();
```
Kingdorian: Same comment as for the CameraHandler.

backend/src/main/java/com/benine/backend/http/PresetsRecallPresetHandler.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Moves the camera
+   * @param camera  A Camera object.
+   * @param preset  The preset to move the camera to.
+   * @throws CameraConnectionException  If the camera cannot be reached.
+   * @throws MalformedURIException      If the request contains an error.
+   */
+  public void moveCamera(Camera camera, Preset preset)
```
Kingdorian: Maybe it would be better to create a preset.apply(Camera c) function but that might be out of scope for this pullrequest.

backend/src/test/java/com/benine/backend/http/CameraStreamHandlerTest.java
```diff
@@ -0,0 +1,112 @@
...
+    doThrow(new IOException()).when(httpresponseMock).getOutputStream();
+    try {
+      when(streamController.getStreamReader(42)).thenReturn(streamReader);
+      streamReaderThread = new Thread(streamReader);
+      streamReaderThread.start();
+    } catch (StreamNotAvailableException e) {
+      e.printStackTrace();
+    }
+  }
+
```
Kingdorian: Could the @Test annotation be missing?

backend/src/main/java/com/benine/backend/http/CameraFocusHandler.java
jochem725: They are not JavaFX imports, but javax imports. This should be correct.

Kingdorian:
Looks good in general but there definitely are some things that need to be addressed in my opinion. Also try to watch out for deleting and replacing files manually next time, this creates a lot of unnecessary things to be marked as changed by github which definitely does not make things more clear.

backend/src/main/java/com/benine/backend/http/CameraFocusHandler.java
Kingdorian: Whoops you're right, it should be fined then.

backend/src/main/java/com/benine/backend/http/CameraStreamHandler.java
jochem725: I think storing it in a file requires it to be read every time. Maybe a better solution would be to create an ENUM or final variables for these headers?

backend/src/main/java/com/benine/backend/http/CameraZoomHandler.java
jochem725: This was some small debug code I forgot to remove. Thanks for pointing it out.

backend/src/main/java/com/benine/backend/http/PresetsRecallPresetHandler.java
jochem725: I do agree with you. I think moving the camera itself should not be a responsibility for the http endpoint. Maybe we should create an issue for this.

backend/src/main/java/com/benine/backend/http/CameraStreamHandler.java
Kingdorian: Well you would only read the file on server creation but a constant would be fine too.

backend/src/test/java/com/benine/backend/http/CameraStreamHandlerTest.java
jochem725: No this is correct, since a thread is needed in most of the tests this is a separate method creating the thread.

jochem725:
@Kingdorian I've addressed all your comments. Something with the files went wrong in my IDE, but I was unable to fix it. I'll be more careful next time. Of course things can be improved such as the way requests are validated now. This can be done in a separate handler for instance.

bvanwijk:
Looks good, however there are some checkstyle warnings:


If these are fixed I agree with merging.

backend/.checkstyle
NaomidR: Checkstyle is already added to gitignore as well.

backend/src/main/java/com/benine/backend/http/CameraFocusHandler.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Created on 21-05-16.
+ */
+public class CameraFocusHandler extends CameraRequestHandler {
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
```
NaomidR: The handle method seems to be very long. Even though it is clear what is done here, maybe this method can be split into smaller methods?

backend/src/main/java/com/benine/backend/http/CameraIrisHandler.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Created on 21-05-16.
+ */
+public class CameraIrisHandler extends CameraRequestHandler {
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
```
NaomidR: Same here, it is pretty long. It doesn't really bother me but I want to have your opinion on why you didn't split this method.

backend/src/main/java/com/benine/backend/http/PresetsRecallPresetHandler.java
```diff
@@ -0,0 +1,76 @@
...
+import com.benine.backend.camera.Position;
+import com.benine.backend.camera.ipcameracontrol.IPCamera;
+import org.eclipse.jetty.server.Request;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class PresetsRecallPresetHandler extends RequestHandler {
```
NaomidR: I think that the name of this class is a bit vague. Can this maybe be changed to something that's more clear?

jochem725: Presets is a prefix for the route it is executed on. Maybe I can make this more clear by renaming and using inheritance.

Kingdorian:
I think you can merge this.

PR #125, opened 2016-05-24
Load from database

egedebruin:
The database now checks if there are the right cameras in the database. If not, changes it in the database.

I have also added some more tests for the database, but I am not able to get the coverage very high. Do you maybe have an idea on how to increase it?

### DIFF
--- a/backend/src/main/java/com/benine/backend/ServerController.java
+++ b/backend/src/main/java/com/benine/backend/ServerController.java
@@ -70,9 +70,9 @@ public class ServerController {
    * Start the server.
    */
   public void start() {
-    startupDatabase();
     cameraController.loadConfigCameras();
-    
+    startupDatabase();
+
     httpController = new HttpController(config.getValue("serverip"),
                         Integer.parseInt(config.getValue("serverport"))); 
     
--- a/backend/src/main/java/com/benine/backend/ServerController.java
+++ b/backend/src/main/java/com/benine/backend/ServerController.java
@@ -130,6 +130,11 @@ public class ServerController {
         e.printStackTrace();
       }
     }
+    try {
+      database.checkCameras();
+    } catch (SQLException e) {
+      e.printStackTrace();
+    }
   }
   
   /**
--- a/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
+++ b/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
@@ -20,7 +20,6 @@ public class MySQLDatabase implements Database {
   private Connection connection;
   private String user;
   private String password;
-  private Logger logger;
   private int presetID;
 
   /**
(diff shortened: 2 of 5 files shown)


backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -4,36 +4,24 @@
...
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
+import java.io.*;
+import java.sql.*;
 import java.util.ArrayList;
 
 /**
- * Created on 4-5-2016.
+ * Created by Ege on 4-5-2016.
```
bvanwijk: We are working as a team so you should replace Ege with Benine or remove your name.

backend/src/main/java/com/benine/backend/ServerController.java
```diff
@@ -130,6 +130,11 @@ private void startupDatabase() {
         e.printStackTrace();
       }
     }
+    try {
+      database.checkCameras();
```
bvanwijk: You are checking the camera's before they are loaded from the config file?
So there are no camera's to compare with yet.
Camera's are loaded from the config in line 74 of this file.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -232,6 +220,97 @@ public void addCamera(int id, String macAddress) {
...
+   * Cheks if there are cameras in the database to be deleted.
+   * @param result The resultset from the query
+   * @param cameras The cameras
+   * @param macs The MACAddresses of the cameras in the database
+   * @throws SQLException No right connection to the database
+   * @throws CameraConnectionException Not able to connect to the camera
+   */
+  public void checkOldCameras(ResultSet result, ArrayList<Camera> cameras, ArrayList<String> macs)
+      throws SQLException, CameraConnectionException {
+    while (result.next()) {
```
bvanwijk: I think it would be better to have a seperate method to retrieve the mac address from the result set.
Then you can just pass only the mac address and the camera's to this method.

backend/src/main/java/com/benine/backend/ServerController.java
egedebruin: Well noticed, will fix this soon.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -232,6 +220,97 @@ public void addCamera(int id, String macAddress) {
...
+
+  @Override
+  public void deleteCamera(int cameraID) throws SQLException {
+    Statement statement = connection.createStatement();
+    try {
+      String sql = "DELETE FROM presets WHERE camera_ID = " + cameraID;
+      statement.executeUpdate(sql);
+      sql = "DELETE FROM camera WHERE ID = " + cameraID;
+      statement.executeUpdate(sql);
+      statement.close();
```
Kingdorian: This can be removed since it will be called in the finally block anyway. Calling close two times might result in weird behaviour.

Kingdorian:
I don't think we should merge this into master ;)

PR #126, opened 2016-05-24
Create tags ui

bvanwijk:
The create preset modal now has a special input for tags.
The available tags list is simulated by a file on the server, because there is no backend supported for yet.


Kingdorian:
Looks good but for me it does not give me suggestions for tags.
image
It seems like the tagnames.json is missing (thats where they're stored right?).
image

bvanwijk:
Nice find, I think you don't have that file because it's included in the gitignore.
I will change it.

frontend/index.html
```diff
@@ -197,10 +209,17 @@ <h4 class="modal-title" id="myModalLabel"><span class="glyphicon glyphicon-wrenc
 
     <!-- HammerJS for supporting touch actions -->
     <script src="bower_components/hammerjs/hammer.min.js"></script>
+	
+	<!-- Bootstrap tagsinput for supporting tags display -->
+	<script src="bower_components/bootstrap-tagsinput/dist/bootstrap-tagsinput.min.js"></script>
```
jochem725: Great find!

frontend/public/css/style.css
```diff
@@ -141,11 +141,15 @@ input[type=range]:focus {
...
 .modal img {
-    position: relative;
-	left: 500px;
-	bottom: 100px;
-    height: 250px;
-    width: 350px;
+    width: 100%;
+}
+.bootstrap-tagsinput {
+	position: absolute;
```
jochem725: I think this positioning causes tags to display weird on some browsers.

bvanwijk: Yes, I would appreciate it, if you can fix that when you are working on the styling part.

Kingdorian:
Working great now! It does not give me suggestions in the tag box for selecting presets on the main screen is it supposed to?
image

bvanwijk:
@Kingdorian I've not implemented it there. Do you think it would be useful to have the same suggestions there? Or would it be better to have the preset image's change on typing every character so you have direct all the presets starting with for example a T or Tro.

Kingdorian:
Maybe even both?

bvanwijk:
@Kingdorian I've added the suggestions to the search box for selecting presets. I think giving the preset images based on the first letters of a keyword is something for in the backend, as the UI is already requesting on every new character the right presets.

jochem725:
I've added some styling to the presets. It should now look correct on all devices.

bvanwijk:
Nice, thank you for styling it. you can merge it.

PR #127, opened 2016-05-24
Load database

egedebruin:
The database now checks if there are the right cameras in the database. If not, changes it in the database.

I have also added some more tests for the database, but I am not able to get the coverage very high. Do you maybe have an idea on how to increase it?

UPDATE:
I have increased the amount of tests, I think the coverage is good now.
I think I have also changed all the things you have commented on, but I needed to close the last pull request because it was merging in master.


bvanwijk:
Nice, that you have managed to increase the test coverage and fixed some comments I had.
I agree with merging.

backend/src/main/java/com/benine/backend/database/Database.java
```diff
@@ -2,11 +2,14 @@
...
 import com.benine.backend.Preset;
 
+import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
 
 /**
  * Interface for communication with the database.
+ *
+ * @author Ege
```
jochem725: Please remove this @author annotation, or change it to the teamname.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -132,7 +121,6 @@ public void updatePreset(Preset preset) throws SQLException {
         list.add(getPresetsFromResultSet(resultset));
       }
       resultset.close();
-      statement.close();
     } finally {
       if (statement != null) {
```
jochem725: I see this piece of code is returning often. Maybe there could be a close(statement) method, which also does the null check.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -148,12 +136,14 @@ public boolean connectToDatabaseServer() {
       String connect = "jdbc:mysql://localhost:3306?useUnicode=true&useJDBCCompliantTimezoneShift="
           + "true&useLegacyDatetimeCode=false&serverTimezone=UTC";
       connection = DriverManager.getConnection(connect, user, password);
-    } catch (Exception e) {
+    } catch (SQLException e) {
+      e.printStackTrace();
```
jochem725: Maybe you could use the logger here to make more clear why the error occurs.

Kingdorian: This seems like an important exception we might want to log this.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -232,12 +221,99 @@ public void addCamera(int id, String macAddress) {
...
+    } finally {
+      if (statement != null) {
+        statement.close();
+        resultset.close();
+      }
+    }
+  }
+
+  /**
+   * Cheks if there are cameras in the database to be deleted.
```
jochem725: There is a small typo here.

backend/src/test/java/com/benine/backend/database/MySQLDatabaseTest.java
```diff
@@ -110,18 +120,134 @@ public final void testGetAllPreset() throws SQLException {
...
+    }
+
+    @Test
+    public final void testCheckNewCameras() throws CameraConnectionException {
+        Camera camera = mock(Camera.class);
+        when(camera.getMacAddress()).thenReturn("mas");
+        database.connectToDatabaseServer();
+        ArrayList<Camera> list = new ArrayList<Camera>();
+        ArrayList<String> macs = new ArrayList<String>();
+        list.add(camera);
```
jochem725: I think the code above this comment could benefit from an @Before method, to reduce the amount of code. Some smaller lines such as `database.connectToDatabaseServer()` also appear in other tests.

Kingdorian:
Test coverage is still at 79% and I can see a lot of bad weather behaviour not being teseted for example in the addCamera and checkCameras methods so I think this can still be improved.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -232,12 +221,99 @@ public void addCamera(int id, String macAddress) {
...
+      checkNewCameras(cameras, macs);
+    } catch (SQLException e) {
+      e.printStackTrace();
+      getLogger().log("Cameras could not be gotten from database.", LogEvent.Type.CRITICAL);
+    } catch (CameraConnectionException e) {
+      e.printStackTrace();
+    } finally {
+      if (statement != null) {
+        statement.close();
+        resultset.close();
```
jochem725: If 'statement == null` the `resultset` stream may not be closed, it this correct behavior?

egedebruin:
I have fixed the things you commented on and increased to test coverage to almost 90%.

bvanwijk:
Looks good now, I agree with merging

jochem725:
Seems fine, I also agree with merging.

PR #128, opened 2016-05-24
Fix travis ipcamera tests

bvanwijk:
This fixes the problems we had with travis killing all the IPcamera tests.


jochem725:
Seems fine! Thanks @bvanwijk for fixing this issue.

egedebruin:
Great fix!

Kingdorian:
Looks good, I think its better to stay consistent with the mocking this way (no mockserver and mockito mixed).

PR #132, opened 2016-05-25
Available tags endpoint

Kingdorian:
The list of tags is now available in the json returned by the the /presets/ endpoint.


backend/src/main/java/com/benine/backend/PresetController.java
```diff
@@ -13,6 +14,8 @@
 
   private ArrayList<Preset> presets = new ArrayList<Preset>();
 
+  private HashSet<String> tags = new HashSet<>();
```
bvanwijk: We also have a static keywords set in the preset class.
I think we have to choose one of them not both.

Kingdorian: I'll remove them from the preset class, I think it's more logical to have them in the preset controller since they're only used for controlling the presets. Do you agree with that?

Kingdorian: This should now be fixed, @bvanwijk.

bvanwijk: Yes, I agree with that nice that you've have fixed it already.

bvanwijk:

jochem725:
Seems fine! I like the refactor you did in `CameraInfoHandler`, as it improves readability.

PR #133, opened 2016-05-25
Keywords in database

egedebruin:
This makes it able to add, delete en get tags to/from the database.

I will create tests for this if the other branch is merged and i can merge from develop. I maked this pull request so you can review it already.

### DIFF
--- a/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
+++ b/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
@@ -144,7 +144,7 @@ public class MySQLDatabase implements Database {
       connection = DriverManager.getConnection(connect, user, password);
       boolean connected = !connection.isClosed();
       return connected;
-    } catch (Exception e) {
+    } catch (SQLException | ClassNotFoundException e) {
       getLogger().log("Connection with database failed.", LogEvent.Type.CRITICAL);
       return false;
     }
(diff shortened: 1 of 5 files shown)


Kingdorian:
I might be mistaken but I don't see any tests for the methods. I think you should add those before we can merge this.

egedebruin:
You are not mistaking, I will add those test when the other branch is merged into develop.

jochem725:
@egedebruin Why will you add the tests when the other branch is merged into develop? Is there a dependency you need?

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -245,6 +246,71 @@ public void useDatabase() throws SQLException {
...
+    try {
+      statement = connection.createStatement();
+      String sql = "DELETE FROM tag WHERE name = " + name;
+      statement.executeUpdate(sql);
+      statement.close();
+    } catch (SQLException e) {
+      e.printStackTrace();
+      getLogger().log("Tag couldn't be deleted.", LogEvent.Type.CRITICAL);
+    } finally {
+      if (statement != null) {
```
jochem725: This is duplicate code which could be in a different method. Also, please don't print the stack trace but use the logger instead.

egedebruin:
I have added the tests for the new features i added, I think it is ready for merging now.

I also took you comment into consideration @jochem725 and fixed it in the whole MySQLDatabase class.

bvanwijk:
@egedebruin There are some findbugs error in MySQLDatabase maybe you can fix them before we merge.

egedebruin:
Thanks for noticing @bvanwijk , I have fixed them.

egedebruin:
I have added that it is possible to add tags to, delete tags from and get tags from the database and it will load the presets on startup.

And the needed tests of course (:

bvanwijk:
Good work Ege, nice that you've fixed all the errors I agree with merging.

jochem725:
Great! I think this can be merged.

PR #134, opened 2016-05-26
MJPEG stream optimization

jochem725:
This should greatly improve the performance of the MJPEG streams.

It now reads the mjpeg in blocks instead of just looping over individual bytes for sending.
Also, now NodeJS stops the connection if the web client has disconnected. On a browser request the streams now properly close.

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -21,7 +21,7 @@
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-server</artifactId>
-            <version>9.0.2.v20130417</version>
+            <version>9.3.9.v20160517</version>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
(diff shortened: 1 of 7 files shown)


NaomidR:
Am I missing something here or is the handle method really 70 lines long? If so, can you maybe split this? It is getting hard to read like this.

bvanwijk:
Seems like a good change, I agree with merging

jochem725:
I've made the handle method smaller to make it more readable.

NaomidR:
Looks good now! I agree with merging this.

jochem725:
I've fixed an additional problem. The only problem left now, that a disconnected stream does not reconnect. This only seems to occur if the server distributing the stream closes the connection.

bvanwijk:
Nice that you have fixed an additional problem, looks good.

egedebruin:
Looks good, I think this can be merged.

PR #135, opened 2016-05-26
Ui settings

jochem725:
I've added a settings window to the UI. I've also added a dependency for cookie management, to make reading and writing the cookies less messy.


bvanwijk:

Kingdorian:
Wouldn't it be better to put the settings button in the top right corner? Possibly even as an emblem of a gear or something like that? It kinda is in an awkward spot at the moment left of the middle.
image

NaomidR:
I agree that the settings button looks a bit lost in the middle. It would fit more with the rest in the top right corner.

jochem725:
@Kingdorian @NaomidR I've moved the settings button to the right, with a cog icon. Also there is now a reset to defaults button. I've changed the naming to `Cancel` and `Save & Close`

Kingdorian:
Great! One last thing before we can merge this is that I still think we should get rid of either the x in the top right corner of the modal or the cancel button but I don't know what the others have to say about this.

bvanwijk:
Looks very good now, yes it it's maybe unnecessary to have both of them.

NaomidR:
It might be unnecessary indeed. I think that the cancel button is easier to use on a tablet than a small x in the corner. Besides this, great changes, I agree with merging.

jochem725:
I think it does not matter to have both the cancel button and the x button. Its part of the official Bootstrap modal and the extra x is not distracting.

bvanwijk:
Yes, I agree it is not distracting and it can't give that much confusion.
I agree with merging.

Kingdorian:
Alright I think you can merge this then.

jochem725:
I'll merge it.

Issue #131, opened 2016-05-25
PresetsControllerTest fails when executed as standalone test


Kingdorian:
The PresetsControllerTest fails when executed as standalone test it succeeds however if it is run as part of the maven test cycle.

bvanwijk:
Fixed in #136

PR #136, opened 2016-05-26
Fix preset

### LINKED ISSUE #129: Presets ID
The preset ID should be set in the presetcontroller, and the createPresetHandler should use this preset ID for saving the preset image. The camera ID in a preset should be set correctly in the createPresetHandler.

bvanwijk:
This fixes #129 and #131


backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -287,7 +282,7 @@ public String createAddSqlQuery(Preset preset) {
     if (preset.isAutoiris()) {
       autoir = 1;
     }
-    return "INSERT INTO presetsdatabase.presets VALUES(" + presetID + ","
+    return "INSERT INTO presetsdatabase.presets VALUES(" + preset.getCameraId() + ","
```
egedebruin: Shouldn't this be preset.getId()?

bvanwijk: Nice find, you are right I will change it

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -54,6 +54,8 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
       getLogger().log(e.getMessage(), LogEvent.Type.WARNING);
       respondFailure(request, res);
     } catch (SQLException e) {
+      e.printStackTrace();
```
NaomidR: Do you need e.printStackTrace() here?

egedebruin:
Thanks for fixing, it looks good now. I agree with merging.

jochem725:
Seems like a good fix. I agree with merging too.

NaomidR:
I agree with merging this now

PR #137, opened 2016-05-26
Add the tags to the preset. 

NaomidR:
As I mentioned earlier, there are still some problems with the tests that I haven't been able to figure out. On develop the test coverage is high. Maybe you can point out some things that could be implemented better.


bvanwijk:
Looks good, however there are 2 checkstyle errors after they are fixed I agree with merging.

Kingdorian:
I agree with Bryan!

PR #139, opened 2016-05-26
Edit presets

NaomidR:
So, should the editPresets handler look something like this? The boolean overwrite false should of course not be static but right now you can only change the tag so that's why I put it in there like this.


backend/src/test/java/com/benine/backend/http/CreatePresetHandlerTest.java
```diff
@@ -70,7 +79,6 @@ public void testCameraID() throws Exception{
 
     getHandler().handle(target, requestMock, httprequestMock, httpresponseMock);
 
-    verify(presetController).addPreset(preset);
```
jochem725: I don't think it is a good idea to delete a test statement to make the test pass.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -41,19 +46,28 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
 
       CameraController cameraController = ServerController.getInstance().getCameraController();
       Camera camera = cameraController.getCameraById(Integer.parseInt(camID));
-
+      String tags= request.getParameter("tags");
+      List<String> tagList = null;
+      if (tags != null) {
```
Kingdorian: You can clean this up by initializing tagList as Arrays.asList("none") and changing it if tags != null.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      tagList = Arrays.asList("none");
+    }
+    
+    int presetID = Integer.parseInt(request.getParameter("presetid"));
+    PresetController presetController = ServerController.getInstance().getPresetController();
+    Preset preset = presetController.getPresetById(presetID);
+    
+    if (overwrite == false) {
+      updateTag(preset, tagList);
```
Kingdorian: So this class only handles editing the tags of a preset? I don't see any code for changing the pan/tilt etc settings.

NaomidR: This should indeed become a class that handles editing presets in general. But for now it only handles editing the tags of a preset (see my explanation of the pr)

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
NaomidR: Fixed!

backend/src/test/java/com/benine/backend/http/CreatePresetHandlerTest.java
NaomidR: I deleted this in my other branch because at the time I thought that the error was in the test, which wasn't correct.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
jochem725: Do we want the value `none` there? Can't we just use an empty list?

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -86,7 +87,7 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
    * @throws StreamNotAvailableException  If the camera does not have a stream.
    * @throws IOException  If the image cannot be written.
    */
-  private void createImage(int cameraID, int presetID) throws
+  private static void createImage(int cameraID, int presetID) throws
```
jochem725: I don't think this method belongs here anymore, since the responsibility of this class is handling the http request, not creating preset images. Maybe we should have a static factory for presets so you can just call Factory.getPreset(<parameters>) or something.
What do the other team members think about this?

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,102 @@
...
+public class EditPresetHandler extends RequestHandler {
+  
+  /**
+   * Constructor for a new editPresetHandler that handles editing a preset.
+   */
+  public EditPresetHandler() {}
+ 
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
```
jochem725: This method is very long and some parts of it can be split in smaller methods.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,102 @@
...
+    String overwriteTag = request.getParameter("overwriteTag");
+    String overwritePosition = request.getParameter("overwriteposition");
+        
+    String camID = request.getParameter("camera");
+    IPCamera ipcam = (IPCamera) ServerController.getInstance().getCameraController()
+        .getCameraById(Integer.parseInt(camID));
+        
+    int presetID = Integer.parseInt(request.getParameter("presetid"));
+    Preset preset = ServerController.getInstance().getPresetController().getPresetById(presetID);
+
```
jochem725: You are calling the `ServerController` instance multiple times, maybe you can call it once and store it in a variable to reduce the amount of duplicate code.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,102 @@
...
+   * @param ipcam                         The camera object. 
+   * @param tagList                       The tags belonging to the preset.
+   * @param presetID                      The id of the preset. 
+   * @throws IOException                  If the image cannot be created.
+   * @throws StreamNotAvailableException  If the camera does not have a stream.
+   * @throws SQLException                 If the preset cannot be written to the database.
+   * @throws CameraConnectionException    If the camera cannot be reached.
+   * @throws MalformedURIException        If there is an error in the request.
+   */
+  public void updatePosition(IPCamera ipcam, List<String> tagList, int presetID) throws
```
jochem725: Maybe we also want an updated preset image here, maybe making it optional using a parameter.

backend/src/test/java/com/benine/backend/http/EditPresetHandlerTest.java
```diff
@@ -0,0 +1,108 @@
...
+import com.benine.backend.PresetController;
+import com.benine.backend.ServerController;
+import com.benine.backend.camera.CameraConnectionException;
+import com.benine.backend.camera.Position;
+import com.benine.backend.camera.ipcameracontrol.IPCamera;
+import com.benine.backend.video.MJPEGStreamReader;
+import com.benine.backend.video.Stream;
+import com.benine.backend.video.StreamNotAvailableException;
+
+public class EditPresetHandlerTest { //extends RequestHandlerTest {
```
jochem725: I don't think you should push commented tests to GitHub.

jochem725:
I've placed some comments. Also I think the EditPresetHandler coverage is too low, but this is due to the commented tests.

backend/src/test/java/com/benine/backend/http/EditPresetHandlerTest.java
NaomidR: It was crashing the whole time due to eclipse so I wanted to check something out by commenting it.

jochem725: Then I think you should solve the issues with Eclipse before pushing or exclude the test file from your commit.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
NaomidR: I am calling the setPreset method that also handles the creation of an image. So if I'm not mistaken a new image is created here as well.

jochem725: @NaomidR Ok, then it should be correct.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
NaomidR: This is fixed. I also made the handle method a bit shorter but it is mostly code that requests certain parameters and calls the update methods.

jochem725:
This already looks better, but I think you should really address the static issue.

NaomidR:
Where/ what do you mean with this? In the CreatePresetHandler?

jochem725:
@NaomidR Yes, I've placed a comment there explaining what I mean. I think it's not a good idea to make methods static just to make them accessible everywhere.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
Kingdorian: We do have a factory for presets in the autopresetcration branch I think we should use that here. You can cherry pick the commits I think.
https://github.com/Kingdorian/Contextproject-BeNine/commit/204fa29b2aee924fd89d9efa071f9ded2006a782
https://github.com/Kingdorian/Contextproject-BeNine/commit/67b4e4d52549f70f1a67f2ad0fbe63bdebe7917e
https://github.com/Kingdorian/Contextproject-BeNine/commit/a3ebfeaf50460f052b0242bfbe164fed22c3b1b0

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+      if (overwriteTag.equals("true")) {
+        updateTag(preset, tagList);
+      }
+      
+      if (overwritePosition.equals("true")) {
+        updatePosition(camID,tagList,presetID);
+      }
+    } catch (MalformedURIException | SQLException | StreamNotAvailableException e) {
+      getLogger().log(e.getMessage(), LogEvent.Type.WARNING);
```
Kingdorian: Plrease use the log(String des, Exception e) method here to keep things consistent.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,101 @@
...
+      respondFailure(request,res);
+    } catch (CameraConnectionException e) {
+      getLogger().log("Cannot connect to camera.", LogEvent.Type.CRITICAL);
```
Kingdorian: Same comment as for line 56.

backend/src/main/java/com/benine/backend/http/PresetsHandler.java
```diff
@@ -33,6 +36,7 @@ public PresetsHandler() {
     addHandler("recallpreset", new RecallPresetHandler());
     addHandler("addtag", new AddTagHandler());
     addHandler("removetag", new RemoveTagHandler());
+    addHandler("editpreset", new EditPresetHandler());
```
Kingdorian: If these things are behind the /preset/... endpoint I think edit would suffice, you would get /preset/edit or something similar. Even better would be /preset/[id]/edit to keep it consistent with the camera endpoints.

backend/src/test/java/com/benine/backend/http/EditPresetHandlerTest.java
Kingdorian: I totally agree with Jochem, I think this is very bad practice.

Kingdorian:
I added some comments which I think should be fixed before we can merge this. Also don't forget to update the wiki page with the api documentation.

backend/src/main/java/com/benine/backend/http/PresetsHandler.java
NaomidR: So you mean that you want it to be called "edit" and not "editpreset"?

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
NaomidR: Fixed!

NaomidR:
@Kingdorian  I updated the wiki page.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -92,7 +96,8 @@ public void updatePosition(String camID, List<String> tagList, int presetID) thr
   MalformedURIException {
     IPCamera ipcam = (IPCamera)control.getCameraController()
         .getCameraById(Integer.parseInt(camID));   
-    Preset newPreset = CreatePresetHandler.setPreset(ipcam, tagList);
+    CreatePresetHandler handler = new CreatePresetHandler();
```
jochem725: I don't think this is behavior we want. The responsibility of the `CreatePresetHandler` is handling the endpoint, not setting presets.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,105 @@
...
+  private ServerController control = ServerController.getInstance();
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
+          throws IOException, ServletException {
+    
+    try {
+      String overwriteTag = request.getParameter("overwriteTag");
+      String overwritePosition = request.getParameter("overwriteposition");
+      String camID = request.getParameter("camera");
```
bvanwijk: Why do you request the cameraID from the client? I think the presetID is enough because when you have the preset you can do preset.getCameraId(), this way the client does not have to send redundant information.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -48,14 +50,13 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
       CameraController cameraController = ServerController.getInstance().getCameraController();
       Camera camera = cameraController.getCameraById(Integer.parseInt(camID));
       String tags = request.getParameter("tags");
-      List<String> tagList = null;
-      
+
+      List<String> tagList = new ArrayList<String>();
```
jochem725: I think the `String` between diamonds can be removed because the compiler will know it is a `String` type.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
NaomidR: You're right, I changed it.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,105 @@
...
+      int presetID = Integer.parseInt(request.getParameter("presetid"));
+      String tags = request.getParameter("tags");
+      
+      Preset preset = control.getPresetController().getPresetById(presetID);
+     
+      List<String> tagList = new ArrayList<String>();
```
jochem725: Same holds here about the `String` between diamonds.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,105 @@
...
+    respondSuccess(request, res);
+    request.setHandled(true);  
+  }
+  
+  /**
+   * Updating the tag only.
+   * @param preset the preset to be changed
+   * @param tagList the tag to be added
+   */
+  public void updateTag(Preset preset, List<String> tagList) {
```
jochem725: This method can be private I think.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -0,0 +1,105 @@
...
+   * @param camID                         The id of the camera object. 
+   * @param tagList                       The tags belonging to the preset.
+   * @param presetID                      The id of the preset. 
+   * @throws IOException                  If the image cannot be created.
+   * @throws StreamNotAvailableException  If the camera does not have a stream.
+   * @throws SQLException                 If the preset cannot be written to the database.
+   * @throws CameraConnectionException    If the camera cannot be reached.
+   * @throws MalformedURIException        If there is an error in the request.
+   */
+  public void updatePosition(String camID, List<String> tagList, int presetID) throws
```
jochem725: This method can also be private I think.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
NaomidR: If I remove <String>  then it gives a warning: ArrayList is a raw type. References to generic type ArrayList<E> should be parameterized. Do you think it is better to just say List = new arraylist and get a warning or use <String> ?

bvanwijk:
I saw you had some problems with the tests, I have added a first test for you maybe you can use it to extend the tests a bit.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
jochem725: Strange, then you should leave it as it is.

Kingdorian: That is because you assign tagList with a static type of List<String> to a new ArrayList. You could also do List<A> taglist = new ArrayList<B> where B is a subclass of A. Why is tagList stored as a list and not an ArrayList?

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
Kingdorian: I agree with Jochem I think you should use a presetfactory here.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -87,7 +87,7 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
    * @throws StreamNotAvailableException  If the camera does not have a stream.
    * @throws IOException  If the image cannot be written.
    */
-  private void createImage(int cameraID, int presetID) throws
+  public void createImage(int cameraID, int presetID) throws
```
jochem725: I think this method can be private.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -132,7 +132,7 @@ public Preset setPreset(IPCamera camera, List<String> tagList)
    * @return          A Preset object.
    * @throws CameraConnectionException If the camera cannot be reached.
    */
-  private Preset createPreset(IPCamera camera, List<String> tagList) 
+  public Preset createPreset(IPCamera camera, List<String> tagList)
```
jochem725: I think this method can also be private.

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -71,34 +81,78 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
...
    * @param presetID                      The id of the preset. 
+   * @param preset                        The preset to be updated.
    * @throws IOException                  If the image cannot be created.
    * @throws StreamNotAvailableException  If the camera does not have a stream.
    * @throws SQLException                 If the preset cannot be written to the database.
    * @throws CameraConnectionException    If the camera cannot be reached.
    * @throws MalformedURIException        If there is an error in the request.
    */
-  public void updatePosition(String camID, List<String> tagList, int presetID) throws 
+  public void updatePosition(Preset preset, int presetID) throws
```
jochem725: This method can be private, and most of the other methods in this class as well.

jochem725:
I think this can be merged now.

bvanwijk:
I'll agree with merging now

bvanwijk:
I've merged develop into this branch, please verify everything is still working correctly.

PR #140, opened 2016-05-26
Preset edit ui

bvanwijk:
The editing is not completly working since the functionality is not yet implemented in develop.
However it is possible to select a preset to edit and see the image of the preset.


jochem725:
Looks fine, I'll handle the styling once both create and edit branches have been merged.

egedebruin:
This looks good and I think this can be merged.

Kingdorian:
Looks good but what is the difference between save and overwrite exactly?

bvanwijk:
@Kingdorian with save you just save the changes you made to the tags and with overwrite you can replace this preset with the current view of the camera.

Kingdorian:
Okay, that clears things up a bit, however it is  not fully clear why we would need two seperate buttons for that. e two? Also once I've selected a camera it does not allow me to switch to another camera.

bvanwijk:
@Kingdorian  How would you prefer to choose between these options then? using a checkbox? Yes, when you have selected the edit icon for editing a camera you must first close the modal before you can select another camera to edit, or is that not what you meant?

Kingdorian:
It does this even when I've already closed the modal. Why can't we update the tags and move the preset (and change other preset attributtes) with one button I these changes often happen at the same time.

bvanwijk:
@Kingdorian that is exactly what the overwrite button will do right now, but I understand this can give some confusion. I don't understand what the other problem is you are talking about?

Kingdorian:
When I click on one of the 4 camera images on the top left quarter of the screen the image on the top right quarter does not change only the number drawn in the top right corner changes.

bvanwijk:
@Kingdorian I'm not able to reproduce your problem, however there is some delay when switching camera's. The reason for this dealy is that all the information about the camera is retrieved on a switch so the controls can give the right information. The information is in the backend retrieved from the camera with several calls so this is why this takes some time. However this is not part of this pull request.  The save button is for just changing the tags for that presets since that is the only information you can currently change in the edit preset window. For now I will change it to just one button. When the implementation is ready for really editing presets we can think about how we want to do the different editings.

Kingdorian:
It is not a delay, I understand why there would be delay but it just does not work anymore, it does work on develop. 

Just changing the button is fine but I think we should find out what the cause of the problem is before we merge this into develop.

Kingdorian:
I've confirmed this issue (now described in #146 )  is not specific to this branch. I think we can merge this

PR #142, opened 2016-05-26
Cameras in use in backend

egedebruin:
You can now get a list of the used cameras and set a camera to used (or not used) in the backend server.


backend/src/main/java/com/benine/backend/camera/BasicCamera.java
```diff
@@ -73,4 +76,14 @@ public boolean equals(Object obj) {
     }
     return false;
   }
+
+  @Override
+  public boolean isInUse() {
+    return inUse;
+  }
+
+  @Override
```
Kingdorian: I think it would be clearer to, setInUse, and setNotInUse or something similar.

Kingdorian:
The code looks good and the testing is nice. I think it would be nice however to keep track of who uses a camera but this might be something to add in a next sprint. It would also be nice to have the inuse time out so that it resets if setInUse has not been called for n (5 or so) seconds. This way if a client crashes or something like that it automatically returns to the not in use state after a couple of seconds.

backend/src/main/java/com/benine/backend/camera/BasicCamera.java
egedebruin: Yes I also thought of that as an option, I will change it.

Kingdorian:
We need to merge #144 first.

jochem725:
After #144 has been merged, I agree with merging.

PR #144, opened 2016-05-26
Camera in use http

Kingdorian:
You can now get info about a camera's status (inuse:true|false) when getting the camera info also a camera can be set to in use or free throught the inuse endpoint.

### DIFF
--- a/backend/src/main/java/com/benine/backend/http/SetCameraInUseHandler.java
+++ b/backend/src/main/java/com/benine/backend/http/SetCameraInUseHandler.java
@@ -22,17 +22,12 @@ public class SetCameraInUseHandler extends CameraRequestHandler {
     int camID = getCameraId(request);
     Camera cam = getCameraController().getCameraById(camID);
 
-    boolean autoOn = Boolean.parseBoolean(request.getParameter("inuse"));
+    String autoOn = request.getParameter("inuse");
 
-    try {
-      if(!cam.isInUse()) {
-        cam.setInUse(autoOn);
-        respondSuccess(request, res);
-      } else {
-        respondFailure(request, res);
-      }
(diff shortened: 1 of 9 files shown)


backend/src/main/java/com/benine/backend/http/SetCameraInUseHandler.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest req, HttpServletResponse res)
+          throws IOException, ServletException {
+    int camID = getCameraId(request);
+    Camera cam = getCameraController().getCameraById(camID);
+
+    String autoOn = request.getParameter("inuse");
+
+    if (isAllowed(cam) && autoOn != null && !cam.isInUse()) {
```
egedebruin: If a client wants to set a camera to not inuse, will it get trough this statement? Because the camera is still in use i think not, and then the inuse cannot be set to false

Kingdorian: Nice find, this went wrong with merging, I'll fix it.

jochem725:
After you fix @egedebruin's comment, I think this can be merged.

bvanwijk:
I agree with Jochem.

Kingdorian:
@egedebruin fixed.

backend/src/main/java/com/benine/backend/http/SetCameraInUseHandler.java
Kingdorian: @egedebruin fixed!

Kingdorian:
@jochem725 @bvanwijk  this should be fixed.

bvanwijk:
I think we can merge it now

PR #145, opened 2016-05-27
Delete tag endpoint

Kingdorian:
Implemented deleting tags via the removetag endpoint.


jochem725:
Looks good, however I think you forgot to add the `removetag` endpoint to the `PresetsHandler`

Kingdorian:
Thanks, I totally forgot that.

Kingdorian:
@jochem725 this should be fixed now.

bvanwijk:

PR #147, opened 2016-05-27
Endpoint add tag

Kingdorian:
This adds an endpoint which can be used to add tags.


backend/src/main/java/com/benine/backend/http/AddTagHandler.java
```diff
@@ -0,0 +1,27 @@
...
+ * Created by BeNine on 25-5-16.
+ */
+public class AddTagHandler extends RequestHandler {
+
+  @Override
+  public void handle(String s, Request request, HttpServletRequest httpServletRequest,
+                     HttpServletResponse httpServletResponse) throws IOException, ServletException {
+    String tagName = request.getParameter("name");
+    if (tagName != null) {
+      ServerController.getInstance().getPresetController().addTag(tagName);
```
jochem725: Do we want to respond success even if the `tagname == null`?

bvanwijk:

jochem725:
Seems fine, I've added a small comment, after that I think this can be merged.

Kingdorian:
@bvanwijk It should be fixed now.

bvanwijk:
Great fix, I think it can be merged now.

PR #149, opened 2016-05-27
Add coveralls support to pom

Kingdorian:
Coveralls should now automatically generate a coverage report for every travis build.  Currently it's configured to make builds with a coverage <85% fail.


jochem725:
Great! This will make it more easy to maintain our coverage, so it stays high!

PR #151, opened 2016-05-27
Fix thread based failing test

jochem725:
This should fix some tests randomly failing.
The thread is no longer needed because the implementation of the handler changed a bit.


bvanwijk:
Nice, that you have fixed the problems we had with randomly failing tests.
I agree with merging.

egedebruin:

PR #152, opened 2016-05-27
Setup coveralls

Kingdorian:
Coveralls should now automatically capture coverage date generated by travis builds. (I accidenally pushed the Coveralls security token but I generated a new one so this one is now invalidated)


Kingdorian:
Something has gone wrong when I tried to remove the token from the history using this guide: https://help.github.com/articles/remove-sensitive-data/. I'll create a new branch and cherry pick the right commits.

PR #155, opened 2016-05-27
Setup coveralls

Kingdorian:
Coveralls should now automatically capture coverage date generated by travis builds. (I accidenally pushed the Coveralls security token but I generated a new one so this one is now invalidated)

https://github.com/Kingdorian/Contextproject-BeNine/pull/152


jochem725:
I've resolved all issues, now Coveralls should correctly receive the test coverage information.

bvanwijk:

egedebruin:
I agree with merging, looks good

PR #156, opened 2016-05-27
Sprintplan 6 and retrospective 5

bvanwijk:
Sprintplan 6 and retrospective 5


egedebruin:
Very good, very mergeable

jochem725:
This looks as agreed on ShareLatex, I think this can be merged.

PR #158, opened 2016-05-31
Databasefix

egedebruin:
@bvanwijk and I fixed some bugs in the database, could you double check there are no errors anymore in the database? Otherwise I think this can be merged.

### DIFF
--- a/backend/configs/main.conf
+++ b/backend/configs/main.conf
@@ -18,3 +18,5 @@ serverip=localhost
 serverport=8888
 sqluser=root
 sqlpassword=root
+
+imagepath=/static/presets/
--- a/backend/src/test/java/com/benine/backend/PresetControllerTest.java
+++ b/backend/src/test/java/com/benine/backend/PresetControllerTest.java
@@ -23,167 +23,141 @@ import static org.mockito.Mockito.when;
 public class PresetControllerTest {
   
   private ServerController serverController;
+  private PresetController presetController;
+  private Preset preset;
+  private Preset preset2;
   
   @Before
-  public void setup() {
+  public void setup() throws SQLException {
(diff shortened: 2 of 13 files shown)


backend/configs/main.conf
```diff
@@ -17,4 +17,6 @@ camera_4_address = 192.168.10.101
 serverip=localhost
 serverport=8888
 sqluser=root
-sqlpassword=root
\ No newline at end of file
+sqlpassword=root
```
Kingdorian: Maybe we should pass these as command line parameters when starting the servers (maybe keep these as default if they're not provided in the command line).

Kingdorian:
Looks good!

backend/src/test/java/com/benine/backend/PresetControllerTest.java
```diff
@@ -43,6 +43,20 @@ public void testAddPreset() throws Exception {
   }
   
   @Test
+  public void testUpdatePreset() throws Exception {
```
jochem725: I think most of the code in these tests can be put in an `@Before` statement. It cleans up a lot of duplicate code in the tests.

jochem725: I think most of the code in these tests can be put in an `@Before` statement. It cleans up a lot of duplicate code in the tests.

jochem725:
I've placed a small comment at a test, after it is fixed I think this can be merged.

NaomidR:
Looks good, I do agree with the comments above though. There are no checkstyle/pmd/findbugs warnings so after fixing the above I agree with merging.

jochem725:
Great! I'll merge this.

PR #159, opened 2016-06-01
Cleanup javascript

### LINKED ISSUE #85: Getcamerainfo returns mallformed json
THe array of camera's is not an array of json data but an array of strings.

bvanwijk:
We now have objects for the camera's en presets in the javascript.
You can now create a preset from a simple camera which only has an image.
I also fixed #85 it now returns the right Json to the client.


egedebruin:
Looks like a very well structured JavaScript code and good cleanup, i agree with merging

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -60,11 +61,15 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
         IPCamera ipcam = (IPCamera) camera;
         setPreset(ipcam, tagList);
         respondSuccess(request, res);
+      } else if (camera instanceof SimpleCamera) {
```
Kingdorian: I think the instanceof is a code smell. You might want to consider moving the important parts of the code within this statement to the SimpleCamera class maybe even create a presetCamera interface or something for a camera that supports presets.

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -120,6 +125,20 @@ private void setPreset(IPCamera camera, List<String> tagList)
...
+   * Creates a preset from a camera.
+   * @param camera    The camera to create the preset from.
+   * @param tagList   The tag belonging to the preset. 
+   * @return          A Preset object.
+   * @throws CameraConnectionException If the camera cannot be reached.
+   */
+  private Preset createSimplePreset(SimpleCamera camera, List<String> tagList) 
+      throws CameraConnectionException {
+    int cameraId = camera.getId();
+    return new Preset(new Position(-1, -1), -1, -1, -1, false, -1,
```
Kingdorian: Maybe we should create a superclass preset with subclasses SimpleCameraPreset and IPCameraPreset.

frontend/public/js/main.js
```diff
@@ -0,0 +1,97 @@
...
+		$('#preset_create_div .tags_input').tagsinput('removeAll');
+	}
+}
+
+/**
+* Search all camera's to find the camera with ID
+* @param id of the camera to search for.
+*/
+function findCameraOnID(id){
+	var res = $.grep(cameras, function(item, n) {
```
jochem725: The n in this function seems to be unused.

jochem725:
Seems fine, I think this can be merged after Dorian's comment is fixed.

bvanwijk:
@Kingdorian I have created a seperate class for simplePresets and IPCameraPresets.

NaomidR:
It looks good that there's now a IPCameraPreset class! I agree with merging this.

backend/database/databasefile.sql
```diff
@@ -53,12 +53,29 @@ CREATE TABLE IF NOT EXISTS `presetsDatabase`.`presets` (
...
 DEFAULT CHARACTER SET = utf8;
 
+-- -----------------------------------------------------
+-- Table `presetsDatabase`.`simplepresets`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `presetsDatabase`.`simplepresets` (
+  `ID` INT(11) NOT NULL,
+  `Image` CHAR(50) NULL DEFAULT '',
+  `camera_ID` INT(11) NOT NULL,
+  PRIMARY KEY (`ID`, `camera_ID`),
```
egedebruin: The primary key is only for the ID, not with the camera_ID. This was already wrong in the presets table, but only now i notice it.

egedebruin:
I've added a small comment, but other than that I again agree with merging. It looks good.

jochem725:
Seems fine, I think you can merge it.

egedebruin:
Looks good, you can merge it I think.

jochem725:
I'll merge it.

PR #160, opened 2016-06-01
Reconnect mjpeg

jochem725:
I've added a reconnection thread to the Stream class.

If a connection drops, or cannot be read, the server will reconnect to that stream, retrying every 5 seconds. You can test this feature by unplugging your internet connection for a minute (or more) and then refreshing the web interface.


bvanwijk:
Looks good, nice to see no checkstyle and findbugs warnings.
However the "tuin camera" camera 1 does not reconnect while the other two camera's do after I disconnected and reconnected the internet.

backend/src/main/java/com/benine/backend/video/Stream.java
```diff
@@ -1,57 +1,132 @@
...
+   * Halts a thread for a specified duration and logs it to the logger.
+   * @param duration  Duration in milliseconds.
    */
-  private InputStream fetchInputStream() throws IOException {
-    return connection.getInputStream();
+  private void wait(int duration) {
+    try {
+      Thread.sleep(duration);
+    } catch (InterruptedException e) {
+      logger.log("Stream thread interrupted while sleeping", LogEvent.Type.INFO);
```
Kingdorian: It might be more usefull to also include the stacktrace (exception) when logging this way you can determine why it was interrupted.

jochem725:
@bvanwijk I have not been able to reproduce the issue with that camera. I'll try though, because it is important that the reconnection mechanism is reliable.

backend/src/main/java/com/benine/backend/video/Stream.java
jochem725: I've now added the exception to the log.

jochem725:
@bvanwijk I've reproduced the issue. You are still dependent on HTTP and the server distributing the tuincam, to detect that you're gone. The bird cams are much faster in detecting this. If you wait a minute, the tuincam should disconnect.

bvanwijk:
@jochem725 yes, you're right after a few minutes the tuincam did appear again.
I think if someone else agrees you can merge it.

egedebruin:
I agree with merging indeed, looks good.

jochem725:
I'll merge it then.

PR #162, opened 2016-06-02
MJPEG Compression

jochem725:
There is now some form of compression involved in distributing streams.
The compressions works by resizing the image, greatly reducing the amount of data sent to the client.

It might be heavy for some computers, thats why I've added a toggle to the config, so you can turn it off if your computer is melting apart.


Kingdorian:
Looks really good.

backend/src/main/java/com/benine/backend/video/MJPEGFrameHeader.java
```diff
@@ -0,0 +1,84 @@
...
+  private String contenttype;
+
+  /**
+   * Creates a frameheader object from a bytestream header.
+   * @param header Header in bytes.
+   */
+  public MJPEGFrameHeader(byte[] header) {
+    String headerString = new String(header, StandardCharsets.UTF_8);
+
+    this.boundary = "--polycast";
```
Kingdorian: I don't think we should use hardcoding like this.

jochem725: I've fixed this by creating final variables.

egedebruin:
Very good clean code and good enhancement, I agree with merging.

bvanwijk:
I think this can be merged.

PR #163, opened 2016-06-02
MJPEG

jochem725:
This branch contains the MJPEG changes for this week, the compression and reconnection parts.

Please verify that everything is still working correctly.


bvanwijk:
Looks very good, everything is still working correctly. It is also nice to see a good test coverage and no checkstyle and findbugs errors.

backend/src/main/java/com/benine/backend/video/MJPEGFrameHeader.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Created on 01-06-16.
+ */
+public class MJPEGFrameHeader {
+
+  private int contentlength;
+  private String boundary;
+  private String contenttype;
+
+  private static final String MJPEG_BOUNDARY = "--polycast";
```
Kingdorian: Could you change this so it's not hardcoded to "--polycast"?

Kingdorian:
Looks really good, I think we can merge this.

backend/src/main/java/com/benine/backend/video/MJPEGFrameHeader.java
jochem725: The MJPEG requires a boundary starting with "--". Each stream already has a boundary, but there is no use in reading that boundary. Just replacing it with any "--BOUNDARY" string reduces the amount of parsing needed per frame. Do you have a suggestion on I can make this not hardcoded?

Kingdorian: Oh I didn't properly understand the usage of it then. I would suggest changing it to "--benine" or "--start[something]" since this product is focussed on polycast but not for their use only.

bvanwijk:
I could not find any frontend changes, maybe you can use resized streams in the client so it's really used in the system.

jochem725:
@bvanwijk In all my excitement about resizing streams, I forgot that. Thanks for pointing it out.

backend/src/main/java/com/benine/backend/video/MJPEGFrameHeader.java
jochem725: @Kingdorian Good point about the naming, I've changed it to `--BOUNDARY`.

bvanwijk:
Maybe someonelse can test if everthing still works correctly?
I am not able to create a preset in the UI.

Kingdorian:

egedebruin:
It works on my machine, so I agree with merging. Looks good.

PR #164, opened 2016-06-02
Fix camera blocking

bvanwijk:
The get methods in the IP camera class save the values for 2 seconds so this will reduce the amount of calls to the camera's.

### DIFF
--- a/backend/src/test/java/com/benine/backend/camera/ipcameracontrol/IpcameraTest.java
+++ b/backend/src/test/java/com/benine/backend/camera/ipcameracontrol/IpcameraTest.java
@@ -171,7 +171,7 @@ public class IpcameraTest {
     assertEquals(camera1.hashCode(), camera2.hashCode());
   }
   
-  @Test
+  @Test(expected = IpcameraConnectionException.class)
   public final void testGetJSONFails() throws CameraConnectionException {
     IPCamera camera = new IPCamera("12");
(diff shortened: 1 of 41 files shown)


jochem725:
This seems like a good fix.

backend/src/test/java/com/benine/backend/camera/ipcameracontrol/IpcameraTest.java
```diff
@@ -84,6 +84,25 @@ public final void testGetPosition() throws CameraConnectionException {
...
+    res = camera.getPosition();
+    assertEquals(0, res.getPan(), 0.000001);
+    assertEquals(180, res.getTilt(), 0.000001);
+  }
+  
+  @Test
+  public final void testGetPositionTwice3Seconds() throws CameraConnectionException, InterruptedException {
+    setCameraBehaviour("APC", "aPC80008000");
+    Position res = camera.getPosition();
+    Thread.sleep(2100);
```
Kingdorian: This slows down testing by a lot if this happens to often.  Maybe you can make the amount of time before it times out an attribute (maybe even get it from the config) so you can set it to a really small amount when testing.

Kingdorian:
Looks good apart from the little comment I made I think this can be merged.

Kingdorian:
Great! I think we can merge this.

PR #165, opened 2016-06-02
Refresh client

bvanwijk:
The server status is now again updated every 10 seconds.
The presets will be updated every 5 seconds.


jochem725:
Great! I think this can be merged.

egedebruin:
Works fine indeed, great enhancement!

PR #166, opened 2016-06-02
Tag modal

egedebruin:
I have created a modal for managing tags, so you can update, delete and add them easily. Please look if this is the right design choice, my sense for web design is not very good (:

I have also made it able to add and delete tags in the backend via the ui, all the way to the database. I think I have included everything and I have tested the things I have included, but I might have forgotten something. So also please look carefully at that.


frontend/public/js/presetcontroller.js
```diff
@@ -312,4 +305,138 @@ function loadPresetEditModal(preset) {
...
 	$('.preset-edit-modal').modal('show');
+}
+
+/**
+* Load the tags modal en fill in the tags.
+*/
+function loadTags() {
+	updatedTags = [];
+	deleteTags = [];
+	newTag({name: 'Trumpet'});
```
bvanwijk: I think this was just for testing purposes?

bvanwijk:
When I add a tag, I can't click on it to edit it directly. I must first save it and load the tag modal again so I can edit it.

egedebruin:
Thanks for the comments @bvanwijk, it is fixed now.

egedebruin:
@bvanwijk the tags are now still available after restarting the backend, please check if this also works for you.

bvanwijk:
@egedebruin nice, you've fixed that. However there is some other problem when I delete a tag the tag is not deleted in the database

frontend/public/js/presetcontroller.js
```diff
@@ -13,7 +15,7 @@ function loadPresets() {
 		}
 		for (var t in obj.tags) {
 			if (localTags.indexOf(obj.tags[t]) === -1) {
-				localTags.push(obj.tags[t]);
+				newTag({name: obj.tags[t]});
```
bvanwijk: Why do you use an object with a name attribute? Can't you simply just use it without an object

frontend/index.html
```diff
@@ -274,6 +275,30 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-cog"></span> Settings</
             </div>
         </div>
     </div>
+	
+	<!-- Defines the modal used for updating tags -->
+	 <div id="tags_update_div" class="modal fade tags-update-modal" tabindex="-1" role="dialog" aria-labelledby="tags-update-modal">
+        <div class="modal-dialog modal-lg">
```
jochem725: I think the tag update modal can be smaller by using `modal-sm` or leaving `modal-lg` out, because it does not need the full assigned width.

jochem725:
Looks like a great addition @egedebruin , however tags don't seem to be working and I'm getting an undefined tag when I'm adding tags outside of the modal.

egedebruin:
@bvanwijk @jochem725 Thanks for the comments, I have fixed what you've addressed, please check if it works also on your machines.

bvanwijk:
Seems to work fine now, however there is 1 findbugs error left.
If this is fixed and some else agrees I think this can be merged

jochem725:
Great! I'll merge it.

PR #167, opened 2016-06-02
Endpoint auto preset

Kingdorian:


jochem725:
Seems fine, maybe you can also add a button in the user interface so we can call the pyramid creator.

bvanwijk:

Kingdorian:
@jochem725 will do.

jochem725:
Seems like a great idea.

egedebruin:

PR #168, opened 2016-06-02
Auto preset creation button

Kingdorian:
Implemented the button as suggested in  #167. 
I cannot get the GLYPHICONS to work but for the rest this should now work for autocreating presets using the button in the interface (we also should look at the positioning of the button maybe @jochem725  can do this).


bvanwijk:
Seems fine, I agree with merging

jochem725:
I've added a glyphicon, I'll agree with merging.

PR #171, opened 2016-06-03
Disable controlling camera

Kingdorian:
When trying send a command to the camera (except for commands that only retrieve info) the camera throws an exception when it is in use.


bvanwijk:
@Kingdorian There are still some PMD and findbugs errors.

backend/src/main/java/com/benine/backend/AutoPresetCreator.java
```diff
@@ -27,17 +25,28 @@
...
+          TimeoutException, IOException, StreamNotAvailableException {
+    if (cam.isBusy()) {
+      throw new CameraBusyException("The camera is busy.", cam.getId());
+    }
+    cam.setBusy(true);
     Position camStartPos = cam.getPosition();
 
     ArrayList<Preset> presets = new ArrayList<Preset>();
     for (ZoomPosition pos : generatePositions(cam)) {
+      cam.setBusy(false);
```
bvanwijk: Wouldn't it be better to set the camera to busy until all presets are made instead of switching it between busy and not busy every time.

jochem725: I just wanted to make the same comment, but Bryan was earlier. I think it also prevents unwanted behavior by not switching between busy and not busy.

Kingdorian:
@bvanwijk If you mean the useless parenthesis, I believe those make the code easier to read but I'll see if I can fix them. The unused variables in IPCamera HOR_FOV and the other should not be removed because they'll be used to tweak the auto creation process in the future.

bvanwijk:
@Kingdorian alright then I think it's not a big problem, also since this merge is not yet in develop.

backend/src/main/java/com/benine/backend/AutoPresetCreator.java
Kingdorian: The problem with this is we would need to make a distinction between interacting with the camera from the AutoPresetCreator and the handlers. Which would imply moving checking if the camera is busy to to the Httphandlers (and other places in the system) where the camera's might be controlled. This clashes with the single responsibility system. It's the camera's responsibility to not allow other commands to be executed when its busy. Also if we extend the busy in the future we could also use it to enforce the single command per camera per 130ms rule using this. Then it would be usefull to check here if the camera is busy, if it is then wait and try again after a little while.

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera.java
```diff
@@ -511,7 +528,7 @@ private String verifyResponse(String response, String expected)
    * Waits until the camera has arrived at a location or the timeout has expired.
    * @param pos The position the camera should be at.
    * @param zoom the zoom of the camera
-   * @param timeout the timeout after which to give up waiting
+   * @param timeout the timeout after which to give up waitingh
```
egedebruin: Very very very small typo (:

backend/src/main/java/com/benine/backend/AutoPresetCreator.java
jochem725: But doesn't this also allow the case in which in-between creation of a preset (while the algorithm is running), the camera is moved by for instance the joystick and blocks the algorithm?

backend/src/main/java/com/benine/backend/camera/ipcameracontrol/IPCamera.java
Kingdorian: Fixed!

backend/src/main/java/com/benine/backend/AutoPresetCreator.java
Kingdorian: This could in theory now happen however this should be fixed after the Http endpoints interact with the camera's in asynchronous way. However the controls should be disabled throughout the whole auto preset creation process.  This was the cleanest method I could think of, do you know a cleaner way?

jochem725: No, I think we should see if this really happens in practice and create an issue if it happens. For now this seems the best solution.

Kingdorian: Can we merge this then? @jochem725

bvanwijk:
I think we can merge it then.

jochem725:
This can be merged.

PR #174, opened 2016-06-04
Ui template

jochem725:
I've refactored the user interface to use a tempting engine.
This makes editing much more easy, and allows reuse of different parts across the web interface.

Please re-run `npm install` and verify that everything is working correctly.


bvanwijk:
Looks good, this makes it a lot cleaner. Everything seems to be working correctly. After someonelse agrees, I think you can merge it.

Kingdorian:
I think this can be merged.

egedebruin:
Good work, I will merge it.

PR #175, opened 2016-06-05
Refactor servercontroller

bvanwijk:
I've changed the way the servercontroller is used, the system is now less dependent on the servercontroller class. There is some problem with the database tests which are failing on Travis only. Maybe someone can take a look why this is the case, because I have no ideas anymore.


egedebruin:
I've set the database log file in resetDatabase to null, now the tests won't fail anymore. We also did not need the log file anyway, so this won't hurt the application in any way.

jochem725:
Looks much better, the ServerController now has a single responsibility.

backend/src/main/java/com/benine/backend/database/DatabaseController.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  private void loadPresets() {
+    PresetController presetController = serverController.getPresetController();
+    try {
+      presetController.addPresets(database.getAllPresets());
+      for (Preset preset : presetController.getPresets()) {
+        preset.addTags(database.getTagsFromPreset(preset));
+      }
+    } catch (SQLException e) {
+      logger.log("Cannot read presets from database", LogEvent.Type.CRITICAL);
```
Kingdorian: Maybe we should log the exception here so the stacktrace is also printed. This could be used to determine the cause of the Exception.

backend/src/main/java/com/benine/backend/ServerController.java
```diff
@@ -47,13 +44,13 @@ private ServerController(String configPath) {
...
 
-    database = loadDatabase();
-
-    cameraController = new CameraController();
+    databaseController = new DatabaseController(this);
+    
+    streamController = new StreamController();
 
-    presetController = new PresetController();
+    cameraController = new CameraController(this);
```
Kingdorian: Why not make Servercontroller static? Then you won't have to pass "this" to all the constructors.

backend/src/main/java/com/benine/backend/camera/CameraController.java
```diff
@@ -25,11 +25,21 @@
...
+  private StreamController streamController;
+  
   /**
    * Constructor of the camera controller which creates a camera factory producer.
+   * @param serverController this cameraController belongs to.
    */
-  public CameraController() {
-    camFactoryProducer = new CameraFactoryProducer();
+  public CameraController(ServerController serverController) {
+    config = serverController.getConfig();
```
Kingdorian: A problem with this can occur when the servercontroller decides to use another config or switch to another logger. This change will then not be reflected in these classes. I don't know how big of a problem this will be. But I think it would be good to make these variables (config, logger etc) and the variables in servercontroller final variables. This will ensure they won't (accidentally) diverge. We'll also need to remove the set methods for these final variables.

Kingdorian:
Looks good apart from the little things I pointed out. I can confirm this solves the issue pointed out by BetterCodeHub.

image

Kingdorian:

jochem725:
Great change, removing `this` also makes the code more readable in my opinion. This can be merged.

PR #176, opened 2016-06-06
Queue of presets in database

egedebruin:
It is now possible to store and retrieve a queue of presets in the database. The queue of presets has a name and id and has a list of the presets belonging to the queue, which are ordered by a sequence.

In my excitement I also added that the tags of a preset will also be added to the preset in the database.


backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -107,6 +208,9 @@ public void addPreset(Preset preset) {
       statement = connection.createStatement();
       String sql = preset.createAddSqlQuery();
       statement.executeUpdate(sql);
+      for (String tag : preset.getTags()) {
```
bvanwijk: You add the tags here to the database, but I think the deletion of presets will not work. When you delete a preset these tags has first to be removed before you can delete the preset because of the foreign key constraints.

egedebruin: Well noticed, I will change that.

bvanwijk: I just realized there is another problem with the tagpresets table. Because we now have simplepresets and presets, but the tagpresets table only reffers to the presets table. This would imply the simplepresets can't have tags which should not be the case.

egedebruin: I will temporarily remove the foreign key constraint, so I can fix that issue in another pull request. I was already doing too much in this pull request.
With removing the foreign key, the program will still work right.

egedebruin:
First #177 needs to be merged in this, where the preset issue will be fixed.

jochem725:
Seems fine, if I understand correctly the order in the database determines the order in the Queue after it is stored? Because I think the front-end needs something (like a number) to order on.

egedebruin:
The order will be determined when the queue is created. The elements are pushed in order and gotten back again in that order.

jochem725:
I think an `ArrayList` is indeed the best choice here. I also see you fixed the CheckStyle issues. I agree with merging now.

backend/src/main/java/com/benine/backend/database/Database.java
```diff
@@ -162,4 +160,55 @@
...
+   */
+  ArrayList<Preset> getPresetsList(int queueID);
+
+  /**
+   * Adds a list of presets to an existing queue.
+   *
+   * @param presets The list which needs to be added
+   * @param queueID The id of the queue
+   */
+  void addPresetsList(ArrayList<Preset> presets, int queueID);
```
NaomidR: I also added a (string) name so that it is easier to use in the UI. Otherwise you will only be able to see a number.

egedebruin: This method only gets the list of presets for the queue. The name is added in addQueue method, where a new queue with an id is added. Maybe it would be more convenient to call this method from the addQueue method, I will work on that.

NaomidR: ohh right, then I did not read it properly. That's fine then!

egedebruin:
Now I have the right objects for loading and adding, I have finshed this and think it is finally ready for merging. The queues are now added to the database, deleted from the database and gotten from the database on startup.

jochem725:
Looks fine, If the others agree, this can be merged.

bvanwijk:
Something went wrong with merging, I've fixed it. I will merge it for you.

PR #177, opened 2016-06-06
Fix preset database

### LINKED ISSUE #169: Preset and simplepreset id
It is now possibile to have a preset and a simplepreset with the same id in the database, this should not be possible for connecting with the tags.

egedebruin:
This will fix #169 
There is now a preset table which connects with the tags and presetlist. The IPpreset and simplepreset tables connect to the camera.


bvanwijk:
Is it correct tags added to a preset are not in the tagpreset table in the database in this branch?

egedebruin:
That shouldn't be correct and I think I know why. I will try to fix it.

egedebruin:
I have fixed it, the tags are now correctly updated. Please check if it still works.

bvanwijk:
I saw the edit preset call was not called from the UI. I have fixed that, now everything is working correctly. If somelse confirms, I think you can merge this.

jochem725:
It seems to be working for me now, I agree with merging.

PR #178, opened 2016-06-07
Fix javascript

bvanwijk:
This solves the problem for presets which are reloaded during searching.
I've also removed some files which aren't used any more.


frontend/public/js/preset.js
```diff
@@ -1,21 +1,22 @@
...
-	this.panspeed = panspeed;
-	this.tiltspeed = tiltspeed;
-	this.autoIris = autoiris;
-	this.image = image;
-	this.id = id;
-	this.img = $('<img src="/api/backend' + image + '" >');
-	this.tags = tags;
-	this.cameraid = cameraid;
+function Preset(object) {
+	this.pan = object.pan;
```
jochem725: Maybe `object` can get a different name, as I think it may be confusing.

frontend/public/js/presetcontroller.js
```diff
@@ -383,7 +383,7 @@ function newTag(val) {
 	tagnames.clearPrefetchCache();
  	tagnames.initialize(true);
 	$.get("/api/backend/presets/addtag?name=" + val, function(data) {
-				console.log("create tag respone: " + data);
+		console.log("create tag respone: " + data);
```
jochem725: I think there is a typo in this line.

jochem725:
Great fixes, I also see you removed the test images 😄

NaomidR:
I see you've fixed Jochems comments. I think it looks good and I agree with merging.

egedebruin:
Good fix, I think this can be merged.

PR #179, opened 2016-06-07
Ui sprint7

jochem725:
This branch contains templates for layout management,
I think it is useful to have this in develop for everyone who wants to make changes to the user interface.


egedebruin:
I've already agreed to this, so I think we can merge this.

bvanwijk:
I agree with Ege, I will merge it.

PR #180, opened 2016-06-07
Preset queue controller

bvanwijk:
The presetQueueController class is used to store the different preset queues in the system.
It will assign an unique id to a preset queue when added and is able to create a json string of all preset queues in the controller.


Kingdorian:
Looks good!

egedebruin:
Looks like a good controller, I agree with merging!

NaomidR:
Looks good indeed, I agree with merging as well.

PR #181, opened 2016-06-08
Class that creates a presetQueue 

NaomidR:
We now have a presetQueue that can store presets in order.


bvanwijk:
Looks good, if someone else agrees I think this can be merged.

backend/src/main/java/com/benine/backend/performance/PresetQueue.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.ArrayList;
+
+
+/**
+ * Class that handles making a queue of presets.
+ * Created by Team BeNine.
+ */
+public class PresetQueue {
+
+  private ArrayList<Preset> queue = new ArrayList<Preset>();
```
jochem725: I think you can remove `Preset` from the diamonds in `new ArrayList<Preset>`

jochem725:
Looks good, I've placed a small comment after this has been fixed this can be merged.

egedebruin:
Good work, I will merge this.

PR #190, opened 2016-06-08
Delete preset

jochem725:
## Today I present: Preset deletion!
Please verify everything works correctly since I had some issues with preset updating.

### DIFF
--- a/frontend/public/js/preset.js
+++ b/frontend/public/js/preset.js
@@ -51,7 +51,7 @@ Preset.prototype = {
 	},
 
 	/**
-	 * Recall this preset.
+	 * Delete this preset.
 	 */
 	delete: function() {
 		$.get("/api/backend/presets/deletepreset?id=" + this.id  , function(data) {});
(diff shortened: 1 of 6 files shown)


frontend/public/js/preset.js
```diff
@@ -49,7 +49,14 @@ Preset.prototype = {
 		switchCurrentView(this.cameraid);
 		console.log("Recall preset id: " + this.id);
 	},
-	
+
+	/**
+	 * Recall this preset.
```
bvanwijk: I think you mean delete this preset

bvanwijk:
Except from the little comment I made, I think it looks good.
Everything seems to be working correctly.

jochem725:
@bvanwijk You are right, thanks for pointing it out. I've fixed it.

jochem725:
I also forgot to add test cases, I've resolved this as well.

egedebruin:
Looks good, it works fine here so I agree with merging.

bvanwijk:
Looks great,  think you can merge this now.

NaomidR:
Looks good now, I agree with merging.

PR #192, opened 2016-06-08
Preset name backend

egedebruin:
I have the name to the preset objects. The database is also updated and the endpoint also add a name to the preset, as well als editing a name.

### DIFF
--- a/backend/src/main/java/com/benine/backend/camera/PresetCamera.java
+++ b/backend/src/main/java/com/benine/backend/camera/PresetCamera.java
@@ -12,7 +12,7 @@ public interface PresetCamera extends Camera {
   /**
    * Creates a preset from a camera.
    * @param tagList   The tag belonging to the preset.
-   * @param name      The name of the name
+   * @param name      The name of the preset
    * @return          A Preset object.
    * @throws CameraConnectionException If the camera cannot be reached.
    */
--- a/backend/src/main/java/com/benine/backend/camera/PresetCamera.java
+++ b/backend/src/main/java/com/benine/backend/camera/PresetCamera.java
@@ -11,10 +11,11 @@ public interface PresetCamera extends Camera {
   
   /**
    * Creates a preset from a camera.
-   * @param tagList   The tag belonging to the preset. 
+   * @param tagList   The tag belonging to the preset.
+   * @param name      The name of the name
    * @return          A Preset object.
    * @throws CameraConnectionException If the camera cannot be reached.
    */
-  Preset createPreset(Set<String> tagList) throws CameraConnectionException;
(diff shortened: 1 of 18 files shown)


backend/src/main/java/com/benine/backend/preset/Preset.java
```diff
@@ -130,6 +141,9 @@ public boolean equals(Object o) {
     if (!tags.equals(preset.getTags())) {
```
jochem725: I think these statements can be combined with an `|`

jochem725:
Great to see this added. I agree with merging.

backend/src/main/java/com/benine/backend/camera/PresetCamera.java
```diff
@@ -11,10 +11,11 @@
   
   /**
    * Creates a preset from a camera.
-   * @param tagList   The tag belonging to the preset. 
+   * @param tagList   The tag belonging to the preset.
+   * @param name      The name of the name
```
bvanwijk: I think you mean the name of the preset

backend/src/main/java/com/benine/backend/http/CreatePresetHandler.java
```diff
@@ -46,14 +46,15 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
       
       Camera camera = getCameraController().getCameraById(Integer.parseInt(camID));
       String tags = request.getParameter("tags");
+      String name = request.getParameter("name");
```
bvanwijk: Maybe you could test here if the name is not null or do we allow a preset without a name?

backend/src/main/java/com/benine/backend/http/EditPresetHandler.java
```diff
@@ -43,23 +43,27 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
...
         tagList = new HashSet<>(Arrays.asList(tags.split("\\s*,\\s*"))); 
       }
       if (overwriteTag.equals("true")) {
         updateTag(preset, tagList);
       }
-      
       if (overwritePosition.equals("true")) {
         updatePosition(preset);
       }
+      if (overwriteName.equals("true")) {
```
bvanwijk: I  don't think we need the overwrite parameter, you could make simply this distinction by checking if the name parameter is null or not.

NaomidR:
When you have fixed Bryans comments I agree with merging this.

egedebruin:
Thank you for the comments @bvanwijk , I think I have fixed them now.

bvanwijk:
Looks great, now I agree with merging.

PR #193, opened 2016-06-09
Subviews

Kingdorian:
I changed so now preset creation is based on subviews (which can later be used for auto preset generation tweaking in the ui). They'll also be used to draw rectangles in the preview so the user can see where the automatically created presets will be.

@bvanwijk could you take a look at the failing test for serverController?

### DIFF
--- a/backend/src/main/java/com/benine/backend/preset/IPCameraPreset.java
+++ b/backend/src/main/java/com/benine/backend/preset/IPCameraPreset.java
@@ -1,18 +1,20 @@
 package com.benine.backend.preset;
 
 import com.benine.backend.ServerController;
-import com.benine.backend.camera.*;
+import com.benine.backend.camera.Camera;
+import com.benine.backend.camera.CameraBusyException;
+import com.benine.backend.camera.CameraConnectionException;
+import com.benine.backend.camera.ZoomPosition;
 import com.benine.backend.camera.ipcameracontrol.IPCamera;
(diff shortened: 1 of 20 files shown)


backend/src/main/java/com/benine/backend/preset/IPCameraPreset.java
```diff
@@ -65,7 +64,7 @@ public IPCameraPreset(IPCamera cam, int panSpeed, int tiltSpeed)
     this.position = new ZoomPosition(cam.getPosition(), cam.getZoom());
     this.focus = cam.getFocusPosition();
     this.iris = cam.getIrisPosition();
-    this.autofocus = cam.isAutoFocusOn();
+    this.autoiris = cam.isAutoFocusOn();
```
bvanwijk: I think this should be this.autoriris = cam.isAutoIrisOn()

bvanwijk:
Except from the comment I made, it looks great.

backend/src/main/java/com/benine/backend/preset/IPCameraPreset.java
Kingdorian: Great find I switched those two around I'll fix it.

jochem725:
Looks like a great feature! I think this can be merged.

egedebruin:
Looks really good, I also think this can be merged.

PR #194, opened 2016-06-09
Fix database not setting the right preset id

bvanwijk:
This fixes the problem with the failing test.


Kingdorian:
Great fix, thanks! I think this can be merged.

egedebruin:
I agree with merging.

jochem725:
Looks fine, I think this can be merged.

PR #195, opened 2016-06-09
Prepare statement

egedebruin:
This ensures that there is no SQL Injection possible anymore. The SQL queries are first prepared before adding the parameters in the queries, ensuring that there is only one query executed.


Kingdorian:
I might have looked over them but  I don't see any tests to ensure SQL injection is not possible anymore. You might want to add those.

egedebruin:
Good thinking, I will add thos tests.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -456,17 +452,24 @@ public void checkNewCameras(ArrayList<Camera> cameras, ArrayList<String> macs)
 
   @Override
   public void deleteCamera(int cameraID) {
-    Statement statement = null;
+    PreparedStatement statement = null;
```
jochem725: This method is becoming quite long now, maybe you can create a method to reduce the amount of code duplication.

jochem725:
I've placed a small comment. If Dorian's comment is fixed, I'll agree with merging.

egedebruin:
I have added the tests for the SQL Injection @Kingdorian

bvanwijk:
I agree with merging.

jochem725:
Looks good, I'll agree with merging.

egedebruin:
First #197 needs to be merged in this.

bvanwijk:
I think this can be merged now.

PR #196, opened 2016-06-09
Search box in the performance edit window

bvanwijk:
I have added searching for presets in the performance edit window


jochem725:
Great, I think this can be merged.

egedebruin:
Looks good and works fine, I agree with merging.

PR #197, opened 2016-06-09
Fix preset adding

egedebruin:
This ensures that the presets will be added correctly to the database.


jochem725:
Seems like a good fix. I think this can be merged.

bvanwijk:
I think this can be merged

PR #198, opened 2016-06-09
Fix presetqueues

bvanwijk:
Solves the issue with the presetQueue handlers and saving preset in the queue.


jochem725:
Seems like a good fix, I think this can be merged.

PR #199, opened 2016-06-09
Preset name ui

bvanwijk:
Preset Names are now showed in the UI.
You can edit them and add them during creation of a preset.
If a preset name is defined it will be shown below a preset in the preset area.


egedebruin:
Looks great, I have nothing to add.

Kingdorian:
I agree with merging this.

jochem725:
Looks like a good addition, I'll merge it.

PR #200, opened 2016-06-09
Auto preset creation modal

Kingdorian:
This adds a modal which can be used to use the auto preset functionality. The implementation of this had to be altered a fair bit to be able to draw the rectangles on the image in the modal.


backend/src/main/java/com/benine/backend/http/PresetsHandler.java
```diff
@@ -63,7 +65,7 @@ public void handle(String s, Request request, HttpServletRequest req, HttpServle
    */
   private String getRoute(Request request) {
     String path = request.getPathInfo();
-
+    System.out.println(path);
```
bvanwijk: I think this can be removed

Kingdorian: Fixed in d3e9588.

bvanwijk:
I think it looks great, I agree with merging.

egedebruin:
Great work, I agree with merging too.

jochem725:
Looks like a great addition. The view is really awesome.

Issue #201, opened 2016-06-09
Too many button colors


jochem725:
We should create one or two basic colors, to make sure only functional buttons have a color (red, green for instance.)

bvanwijk:
fixed in #214

PR #202, opened 2016-06-09
Preset Performance Lists

jochem725:
I've added performances to the ui.

You can create a performance in the edit window.
You can add presets by searching them. You can move them around, ordering them. 

In the live view, you can open a performance. If the preset is 'current' it is automatically recalled, if a preset is 'up next' it is the next preset to be automatically recalled.

A known issue is the high amount of colored buttons, I've created #201 to make sure we address this in the future. Also, currently no tags are loaded in the live view queue.


backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -190,7 +191,7 @@ public void addQueue(PresetQueue queue) {
       statement = connection.prepareStatement(sql);
       statement.setInt(1, queue.getID());
       statement.setString(2, queue.getName());
-      statement.executeUpdate();
+      statement.executeUpdate(sql);
```
egedebruin: When you execute a prepared statement, you do not need to re-assign the query in de execution.

frontend/public/js/editperformance.js
```diff
@@ -0,0 +1,292 @@
...
+
+    drawSchedule(selectedPerformance.data());
+}
+
+function setNameEditable() {
+    if (selectedPerformance !== undefined) {
+        var element = selectedPerformance;
+
+        var li = $('<li><div class="row">' +
+            '<div class="col-xs-7"><input type="text" id="performance-name" class="form-control" value="' + $(element).data()['name'] + '"></div> ' +
```
Kingdorian: Can't we do something cleaner (and more fancy) with ejs here?

Kingdorian:
Except for lots of "hardcoded" html in the javascript which we might want to replace with some ejs stuff (possibly in the future) the code looks really good and layout very clean!

bvanwijk:
Looks great, nice new feature I think we can merg it.

frontend/public/js/editperformance.js
jochem725: Since ejs is rendered in nodeJS and not on the client this is not possible.
However I agree with you that in the future the code can be refactored to be come more clean.

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
jochem725: Thanks for pointing that out, I'll fix it.

egedebruin:
The database kept on resetting on my machine, I have now fixed that together with you (:
I agree with merging now.

jochem725:
Great, thanks @egedebruin!

PR #209, opened 2016-06-10
Cleanup warnings

bvanwijk:
I fixed some small checkstyle and findbugs warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ jspm_packages
 #Images added for presets
 backend/static/presets/*
 !backend/static/presets/1_0.jpg
+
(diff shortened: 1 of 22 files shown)


jochem725:
Seems fine, thanks for fixing this @bvanwijk. Maybe we should mention somewhere why the FindBugs warning was excluded.

bvanwijk:
@jochem725 it is mentioned in the findbugs-exclude.xml file and above the method where the warning occurs.

jochem725:
@bvanwijk Oh I missed that, then I think we can merge this.

NaomidR:
Looks fine, this can be merged.

PR #212, opened 2016-06-12
Camera inuse

bvanwijk:
It is now possible to see in the UI which camera's are in use by other users.


jochem725:
Great to see this working, I agree with merging.

Kingdorian:
Looks good, I would move the timeout setting to the config file however. Also 5000ms might be a bit short since the auto preset creation sometimes spends more time waiting then this.

bvanwijk:
@Kingdorian I don't completely understand what the auto preset creation has to do with this. The clients say to the backend every few seconds which camera they have in use. During the auto preset creation the client will keep doing this so the others know which camera is in use. This does not block other clients to use the camera. 
Do you have a suggestion how to reach the config from the basicCamera class since a static call to the serverController seems not like a good idea to me?

Kingdorian:
Well that way if the camera is auto creating and for some reason it drops the camera will be released but the auto preset creation process is still going on since the client does not actively interact with this. It does however interact with the camera's to say they're busy every once in a while (for now it is 6 seconds but this could be shortened). 

It might not be the most elegant method but you could create a new config object in the basicCamera class. But on second thought it does not add a lot of value to have this in the config maybe. If it will not be retrieved from the config I would suggest changing the name of _useTimeOut_ to upper case to indicate its a constant.

bvanwijk:
@Kingdorian I have changed useTimeOut to upper case.

Kingdorian:
@bvanwijk  great but the autopreset creation issue when clients disconnect is still not solved I don't think we should merge this before it is.

bvanwijk:
@Kingdorian What do you propse as a solution how many seconds must the timeout be?

bvanwijk:
@Kingdorian I think I have fixed it now, before auto creating every preset the camera is set in use for 7 seconds. The auto creating of 1 preset takes around 6 seconds after which the camera is set in use again. So the camera stays in use during the auto preset creation even if the client crashes.

Kingdorian:

jochem725:
If someone else agrees, this can be merged.

NaomidR:
I agree with merging this

PR #213, opened 2016-06-12
Image creation

bvanwijk:
Preset images are now created at one place in the code.
When a preset is removed from the system, the image will also be removed.

### DIFF
--- a/backend/src/main/java/com/benine/backend/preset/Preset.java
+++ b/backend/src/main/java/com/benine/backend/preset/Preset.java
@@ -1,18 +1,33 @@
 package com.benine.backend.preset;
 
+import com.benine.backend.ServerController;
 import com.benine.backend.camera.Camera;
 import com.benine.backend.camera.CameraBusyException;
 import com.benine.backend.camera.CameraConnectionException;
+import com.benine.backend.video.MJPEGFrameResizer;
+import com.benine.backend.video.StreamNotAvailableException;
+import com.benine.backend.video.StreamReader;
+import com.benine.backend.video.VideoFrame;
+
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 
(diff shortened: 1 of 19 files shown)


backend/src/main/java/com/benine/backend/preset/Preset.java
```diff
@@ -157,6 +170,47 @@ public boolean equals(Object o) {
...
+   * @throws IOException  If the image cannot be written.
+   * @throws SQLException if the image can not be saved in the database.
+   */
+  public void createImage(StreamReader streamReader, String folder) throws
+          StreamNotAvailableException, IOException, SQLException {
+
+    File path = getNewImagePath(folder);
+
+    VideoFrame snapShot = streamReader.getSnapShot();
+    MJPEGFrameResizer resizer = new MJPEGFrameResizer(160, 90);
```
jochem725: Group goto-fail asked if we could adjust the snapshot size to be larger, maybe we can change this to `(640,360)`.

jochem725:
Looks much better, I think we can merge this.

backend/src/main/java/com/benine/backend/preset/Preset.java
bvanwijk: @jochem725 I've changed the image size to 640 x 360, but I have also added the image size to the config so it's now easy to adjust.

jochem725:
@bvanwijk Great, I like the config option.

Kingdorian:
Looks really good. But it might be good to add a comment to the config file in the lines above the height and width specifing what the range can be. I get an exception when I enter a negative or a big number.

NaomidR:
Looks good! Cleaner than it was before!

bvanwijk:
@Kingdorian I have added comments about the range in the config file

Kingdorian:
Great, I think we can merge this now.

PR #214, opened 2016-06-13
UI Colors

jochem725:
I've replaced all non-functional button colors with a less distracting color.
Only buttons which need a functional color (toggles, deletion buttons etc.) still have their original color.

### DIFF
--- a/frontend/public/css/style.css
+++ b/frontend/public/css/style.css
@@ -667,3 +667,8 @@ input[type=range]:focus {
     background-color: #3D525B;
     border-color: #4d656f;
 }
+
+.label-info, .progress-bar {
+    background-color: #4d656f;
+
+}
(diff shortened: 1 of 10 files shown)


bvanwijk:
Looks much better, I agree with merging.

Kingdorian:
Looks much less messy. But the buttons for the modals might still be a bit to colorfull.
image
image

jochem725:
@Kingdorian I agree with you, maybe making the close buttons gray instead of green would be better. What do you propose?

NaomidR:
I agree with @Kingdorian about the buttons in the modals, they don't really have a functionality. Gray would be better indeed. After you changed this, I agree with merging it.

Kingdorian:
@jochem725 Yes (blue\ish) gray would be better.

jochem725:
I've now made all green buttons gray.

egedebruin:
Is it correct that the edit preset button should be red?

jochem725:
@egedebruin Yes, in this branch the color is functional (editing, not editing). In #230 this color is removed and replaced with a toggle button.

egedebruin:
Ok, then I agree with merging.

PR #216, opened 2016-06-13
Deployable

bvanwijk:
There is now a zip file generated containing the executable for the backend server and the frontend server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ Build the back-end server using `maven install` with the supplied `pom.xml` file
 
 ### Deployable zip
 Build the back-end server using `maven install` with the supplied `pom.xml` file. 
-The zip file containing the backend and frontend server can be found in the `backend/target/site` folder.
-The backend and frontend server can both be started using the startserver file. 
+The zip file containing the backend and frontend server can be found in the `backend/target` folder.
+When unzipped the backend and frontend server can both be started using one of the startserver scripts, the batch file (.cmd) or the bash script (.sh).
 
 ## Tool results
 ### JavaScript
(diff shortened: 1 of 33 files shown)


Kingdorian:
Maybe it would be good to add to the readme how to create a deployable. I think we should also include a startserver .sh file for unix systems instead of the .cmd.

bvanwijk:
@Kingdorian I agree, but I think it would be better to ask somebody with a unix system to create this.

Kingdorian:
Hahah that's true, I'll add it tomorrow.

Kingdorian:
@bvanwijk Should work now, can someone with an unix system @jochem725  confirm the bash script works? Also is the .cmd script verified to work by someone with a windows system like @NaomidR or @egedebruin ?

jochem725:
I've resolved an issue with the decimal format locale. Please verify if it is now working on all system languages. @Kingdorian I think the .sh file needs changed permissions (`chmod -x` if I'm correct) before it becomes executable.

egedebruin:
The .cmd script works for me.

Kingdorian:
@jochem725 I'm aware that the .sh file needs execute permissions, this can ofcourse not be done automatically but I'll add it to the readme.

jochem725:
@Kingdorian I believe it is actually possible using the `fileMode` configuration, described [here](http://maven.apache.org/plugins/maven-assembly-plugin/assembly.html#sources).
I think adding `<fileMode>0755</fileMode>` below `</includes>` should fix it.

jochem725:
@Kingdorian I've implemented this, and now the .sh file is automatically given execute permissions. I think this can be merged now.

bvanwijk:
I think everything works correctly now, the next person can merge it.

PR #217, opened 2016-06-13
 Duplicate tags

bvanwijk:
Duplicate tags could not be added in the UI anymore.
If also limited the amount of characters for names and tags.
When a tag doesn't have a name the name: preset ID is set for this preset.

### DIFF
--- a/frontend/public/js/presetcontroller.js
+++ b/frontend/public/js/presetcontroller.js
@@ -403,12 +403,14 @@ function updateTags() {
 * @param val value of the tag to add.
 */
 function newTag(val) {
-	localTags.push(val);
-	tagnames.clearPrefetchCache();
- 	tagnames.initialize(true);
-	$.get("/api/backend/presets/addtag?name=" + val, function(data) {
-		console.log("create tag response: " + data);
-	}).done();
+	if (localTags.indexOf(val) < 0 && val != "" && val != undefined) {
+		localTags.push(val);
+		tagnames.clearPrefetchCache();
+		tagnames.initialize(true);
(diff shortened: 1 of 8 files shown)


NaomidR:
Great changes! I do however think that the maximum size for the tags is not enough. If you want to create a bit more complex names, the length can easily be more than 15.

jochem725:
@NaomidR I don't think the point of a tag is to be complex. If the size becomes larger than 15, user interface spacing issues may arise.

jochem725:
Looks great @bvanwijk I agree with merging.

NaomidR:
@jochem725 True, that's what I thought as well but we talked about this and you (I believe) were the one that talked about having more complex/longer (tag) names

jochem725:
@NaomidR No, I was talking about spaces in tags, not complex and longer names.

frontend/public/js/presetcontroller.js
```diff
@@ -403,12 +410,14 @@ function updateTags() {
 * @param val value of the tag to add.
 */
 function newTag(val) {
```
egedebruin: Is it maybe possible to add a return to this function, so that it returns true when it is correctly pushed to the tags (false otherwise)? Then you can know that it is correctly pushed when you call this method.

egedebruin:
I commented one little thing, but other than that I agree with merging.

bvanwijk:
@egedebruin I've added your suggestion.

NaomidR:
What I said before was a misunderstanding. I agree with merging this

PR #218, opened 2016-06-13
settings view

bvanwijk:
Start settings with 1 and increate size of sliders

### DIFF
--- a/frontend/public/css/style.css
+++ b/frontend/public/css/style.css
@@ -131,7 +131,7 @@ input[type="range"]::-ms-tooltip {
 }
 
 .settings-modal .sense_block {
-	width: 100px;
+	width: 110px;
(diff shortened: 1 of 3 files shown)


Kingdorian:
Looks great I think this can be merged.

NaomidR:
Looks good indeed, I agree with merging.

jochem725:
It seems to be bugging for me on 2 digit number values.
schermafbeelding 2016-06-14 om 11 22 21

egedebruin:
It works fine for me, I don't have the same problem as Jochem.

bvanwijk:
@jochem725 I think I've fixed the problem could you verify this?

jochem725:
@bvanwijk It is indeed fixed, I think this can be merged now.

PR #219, opened 2016-06-14
Add new tag to front of list of tags

egedebruin:
This adds a new tag to the top of the taglist, not on the bottom.


bvanwijk:
Seems to work, I agree with merging.

Kingdorian:
I agree with merging this.

NaomidR:
I think this can be merged.

PR #220, opened 2016-06-14
Other name to newly added tag

egedebruin:
When creating a new tag, the default first was "new", but it is changed to "tag" + index of new tag. I think this is better then first, but maybe someone has a better idea.


Kingdorian:
I think we can merge this.

bvanwijk:
Looks good, maybe you could add a space between tag and the index.

bvanwijk:
I agree with merging now.

NaomidR:
I don't have anything to add. I think we can merge this!

PR #221, opened 2016-06-14
auto preset creation input for tweaking

Kingdorian:
There are now input fields for tweaking the amount of rows,  columns and levels for the auto creation process.
image

### DIFF
--- a/frontend/public/css/style.css
+++ b/frontend/public/css/style.css
@@ -660,3 +660,7 @@ input[type=range]:focus {
 #autopreset-left-column {
   padding: 15px;
 }
+
+.label-auto-preset-input {
+  width: 45%;
+}
--- a/frontend/public/css/style.css
+++ b/frontend/public/css/style.css
@@ -109,21 +109,30 @@ nav img {
 
     border-radius: 8px;
 }
+.auto-presets-modal input[type=range] {
+  margin-right: 15px;
+}
+
+.horizontal-slider {
+	z-index: 0;
+  display: block;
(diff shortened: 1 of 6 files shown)


frontend/public/js/autopresets.js
```diff
@@ -1,13 +1,62 @@
+var columns = 3;
```
bvanwijk: It would be nice to have some comments describing what each function does.

frontend/public/css/style.css
```diff
@@ -109,21 +109,30 @@ input[type=range] {
 
     border-radius: 8px;
 }
+.auto-presets-modal input[type=range] {
```
Kingdorian: They where there since this was initially the idea but I decided to replace them with the -/+ buttons as we have now. We will however add a slider for the overlap so that's why I left the styling for them here. The overlap to the UI is another sprint item thats why I didn't include this in this pullrequest.

bvanwijk: Alright, then it's no problem to leave them.

frontend/public/js/autopresets.js
Kingdorian: I added some comments and cleaned up the script a bit in 14cff5cb6f30340b61d71f36b64960a5618ef33f

bvanwijk:
Seems fine, I agree with merging.

egedebruin:
Looks fine, I think this can be merged.

jochem725:
Looks great! I think you can merge this.

PR #224, opened 2016-06-14
Disable auto presetcreation when no camera is selected

### LINKED ISSUE #207: Modal autopresetcreation available when no camera is selected.

bvanwijk:
Fix #207


egedebruin:
Looks fine, good small fix.

Kingdorian:
Looks good, maybe we also want to disable this button if the camera is in use by someone else.

bvanwijk:
@Kingdorian Yes, that would be nice, but the problem with this is the backend doesn't supply information about which client has the camera in use. So you don't no if you're the client with the camera in use or another client has the camera in use.

Kingdorian:
We could possibly add that but I'll create an issue for that since its outside the scope of this branch/sprint item.

jochem725:
I agree with Dorian, I think you can merge this now 😄

PR #225, opened 2016-06-14
Direct add tag

egedebruin:
This ensures that the tags are directly added, deleted and updated, instead of only when clicking on save&close.
This also ensures that the tags are correctly updated in the presets, when deleting and updating tags.


Kingdorian:

jochem725:
Looks fine, I agree with merging.

bvanwijk:
When I add a preset with a tag, then I change this tag the tag in the preset is updated correctly but when I reload the update tags window the old tag and the new tag are both visible.
Maybe you could take a look at this?

egedebruin:
@bvanwijk, I think with the latest commit that should be fixed.

bvanwijk:
I changed a small thing, I think you can merge it now.

NaomidR:
Looks good, I agree with merging this.

PR #226, opened 2016-06-14
Small unit interfaces

bvanwijk:
I have reduced the amount of parameters of some methods as suggested by the bettercodehub tool from SIG.

### DIFF
--- a/backend/src/main/java/com/benine/backend/camera/SimpleCamera.java
+++ b/backend/src/main/java/com/benine/backend/camera/SimpleCamera.java
@@ -55,7 +55,9 @@ public class SimpleCamera extends BasicCamera implements PresetCamera {
   @Override
   public Preset createPreset(Set<String> tagList, String name) throws CameraConnectionException {
     int cameraId = getId();
-    return new SimplePreset(cameraId, tagList, name);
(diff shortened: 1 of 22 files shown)


Kingdorian:
I think the number of parameters in the IPCamera preset still might be a bit to high.

bvanwijk:
@Kingdorian yes, it's still a bit high. What do you think is the best solution for this? Adding the autofocus, autoiris, focus, iris to the position class? or create a seperate class IPCameraPosition with these values?

Kingdorian:
Maybe combine the autoIris boolean and value into a single object the same can be done for the autofocus. This reduces it already with 2. I think the id can be removed completely since this is reassigned after creation of the preset by the preset controller anyway but I might be wrong.

bvanwijk:
The preset id is not in the constructor of a preset, it's the cameraID which is in the constructor.
Sounds like a good idea for the iris and focus, I will implemented that.

jochem725:
I agree with Dorian about the iris and focus parameters. After this is fixed, I agree with merging.

Kingdorian:
This indeed fixes the bettercodehub warning. 
image

jochem725:
Let's merge it then.

PR #227, opened 2016-06-14
Change database file

egedebruin:
This changes the database file so that the database is fully connected.

PR #228, opened 2016-06-14
Database code change

egedebruin:
The code in the database class is now changed to handle with the new database.

It doesn't work properly yet, first #227 and this need to be merged in database-sprint-8, then we can test if the functionality is correct.


Kingdorian:
Looks good.

jochem725:

bvanwijk:
I agree with merging.

PR #230, opened 2016-06-14
Toggle buttons

jochem725:
I've added toggle buttons to the UI.
Please run `npm install` before testing.


bvanwijk:
Looks really good, I agree with merging.

NaomidR:
Much better than the previous button! I agree with merging.

egedebruin:
The buttons look nice, I think this can be merged.

PR #232, opened 2016-06-14
Fix Stream getting stuck.

jochem725:
I think this might fix the streams getting stuck.

Since the logger does not like asynchronous writing, we might want to consider disabling logging http requests to prevent a high amount of exceptions from occurring, or catch them without writing to the logger.


bvanwijk:
Looks like good changes, I agree with merging.

egedebruin:
Looks good, I think this can be merged.

PR #233, opened 2016-06-15
Aspectratio auto preset subviews

Kingdorian:
The generation of the subviews now accounts for the aspect ratio of the camera. It will always try to fit the largest possible subview (with the correct aspect ratio).


bvanwijk:
Seems fine, I agree with merging this.

NaomidR:

Issue #206, opened 2016-06-10
Logger displays IndexOutOfBoundsException


jochem725:
The logger randomly (?) displays an `IndexOutOfBoundsException` which I cannot trace. 
Maybe we can take a look at it.

bvanwijk:
![2016-06-12](https://cloud.githubusercontent.com/assets/9803346/15991343/b31a2d50-30b0-11e6-9804-e4e0d900ef3e.png)

Kingdorian:
I don't know it for sure but I think its when multiple threads (in the http handlers) interact with the logger at the same time. I'll have to take a look at this.

Kingdorian:
Fixed in https://github.com/Kingdorian/Contextproject-BeNine/pull/234

PR #234, opened 2016-06-15
Add semaphore for flushing

Kingdorian:
Fixes https://github.com/Kingdorian/Contextproject-BeNine/issues/206 I cannot replicate the behaviour described in the issue with this fix, can someone confirm this?

### DIFF
--- a/backend/src/main/java/com/benine/backend/LogWriter.java
+++ b/backend/src/main/java/com/benine/backend/LogWriter.java
@@ -30,6 +30,9 @@ public class LogWriter {
 
   private volatile ArrayList<LogEvent> buffer = new ArrayList<LogEvent>();
 
+  // Semaphore for flushing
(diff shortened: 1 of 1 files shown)


bvanwijk:
I've tested it, and I didn't get the issue. I agree with merging.

jochem725:
After running it for over 25 minutes, it appears a nullpointerexception is still possible. 
schermafbeelding 2016-06-15 om 12 26 29

jochem725:
In a second run, the same exception appeared somewhere near 100 minutes.

Kingdorian:
@jochem725 @bvanwijk  I changed this a bit so it should now not happen anymore. I tested it for 2 hours on my laptop with 3 clients and did not get an exception.

jochem725:
Excellent, let's merge it.

PR #235, opened 2016-06-15
Database sprint 8

egedebruin:
This contains everything changed from the database in this sprint, but now we can test the functionality. I have tested some things and now everything works for me, but it is good if you all could double check if everything which has to do with the database still works correctly.

For this the database first needs to be reset on your machines.


bvanwijk:
It seems to work for me, however there are some checkstyle and findbugs warnings which could be fixed.

jochem725:
I can also confirm this seems to be working correctly. I agree with merging.

PR #236, opened 2016-06-15
Update also the image when refreshing presets

bvanwijk:
When the preset image is changed or added later, the image will be loaded in the UI.


jochem725:
Seems like a good fix, I agree with merging.

egedebruin:
Good fix, I also agree with merging.

PR #237, opened 2016-06-15
Change adding presets when auto-creating

NaomidR:
The presets are still automatically added but now a list with the id's of the presets is given back as a response to the client. The client will then be able to say which presets it wants to keep and the rest will be removed (this removing partis part of another sprint task).


backend/src/main/java/com/benine/backend/preset/autopresetcreation/AutoPresetCreator.java
```diff
@@ -16,61 +16,65 @@
...
     cam.setBusy(true);
     ArrayList<IPCameraPreset> presets = new ArrayList<>();
+    ArrayList<Integer> presetIDs = new ArrayList<>();
 
     cam.setBusy(false);
     cam.setAutoFocusOn(true);
     cam.setBusy(true);
     for (ZoomPosition pos : generatePositions(cam, subViews)) {
+      presetIDs.add(generatePresetFromPos(pos,cam).getId());
       presets.add(generatePresetFromPos(pos, cam));
```
jochem725: I see here a preset is first created, then an id is requested and added to the `presetIDs` array. After this, another preset is created. Can't the preset be generated just once instead of twice?

NaomidR: Good point, I changed it.

backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
```diff
@@ -42,9 +45,10 @@ public void handle(String s, Request request, HttpServletRequest httpServletRequ
     IPCamera ipcam = (IPCamera) cam;
    
     try {
-      creator.createPresets(ipcam, creator.generateSubViews());
-      respondSuccess(request, httpServletResponse);
-
+      ArrayList<Integer> presetIDs = creator.createPresets(ipcam, creator.generateSubViews());
+      String presetIDList = presetIDs.toString();
+      respond(request, httpServletResponse, presetIDList);
```
bvanwijk: I think it would be better to have this list of preset IDs in JSON format instead of just the tostring of the arraylist

jochem725: I agree with Bryan on this, JavaScript will have a hard time reading this string representation.

NaomidR: It is now in JSON format

bvanwijk:
I agree with merging now.

jochem725:
I think this can now be merged.

backend/src/main/java/com/benine/backend/preset/autopresetcreation/AutoPresetCreator.java
```diff
@@ -15,62 +15,68 @@
...
-   * @throws InterruptedException when interupted while waiting for cam to move.
+   * @throws InterruptedException when interrupted while waiting for cam to move.
    * @throws TimeoutException if the camera is moving too slow or not at all.
    * @throws CameraBusyException if the camera is busy.
    * @throws IOException if exception occurs when creating the preset image.
    * @throws StreamNotAvailableException if a camera stream cannot be reached.
    * @throws SQLException if exception occurs while writing to the database.
    */
-  public Collection<IPCameraPreset> createPresets(IPCamera cam, Collection<SubView> subViews)
+  public ArrayList<Integer> createPresets(IPCamera cam, Collection<SubView> subViews)
```
Kingdorian: Why is this changed to ArrayList? I think we should keep this an collection.

backend/src/main/java/com/benine/backend/preset/autopresetcreation/AutoPresetCreator.java
```diff
@@ -15,62 +15,68 @@
...
     cam.setBusy(true);
     for (ZoomPosition pos : generatePositions(cam, subViews)) {
-      presets.add(generatePresetFromPos(pos, cam));
+      IPCameraPreset currentPreset = generatePresetFromPos(pos,cam);
+      presetIDs.add(currentPreset.getId());
+      presets.add(currentPreset);
     }
     cam.setBusy(false);
-    return presets;
+    return presetIDs;
```
Kingdorian: I don't think its a good idea to return an list of presetids instead of a collection of preset here since this won't be useful when implementing the progress bar functionality.

backend/src/main/java/com/benine/backend/preset/autopresetcreation/AutoPresetCreator.java
```diff
@@ -15,62 +15,68 @@
...
+ 
   /**
    * Automatically creates presets for the selected camera.
    * Calls generatePositions to get the positions of the presets.
    * Automatically adds the presets to the presetcontroller.
    * @param cam the camera to create presets for.
-   * @param subViews the subviews to generate positions for.
-   * @return A collection of the created presets.
+   * @param subViews the sub-views to generate positions for.
+   * @return A collection of the id's of the created presets.
```
Kingdorian: I think id's should be ids since id's implies possesion.

bvanwijk:

egedebruin:
Looks fine, I agree with merging.

jochem725:
I agree, you can merge it.

PR #238, opened 2016-06-15
Fix wrong intial values tweak parameters autopreset

Kingdorian:
The values indicating the amount of rows, columns and levels would show 2 when the actual value was 3 this updated after pressing a button. This fix should make sure they're correct when you open the modal.


bvanwijk:
Seems like a good fix.

jochem725:
I think this can be merged.

PR #239, opened 2016-06-15
Add install script

Kingdorian:
I added a script that can be used for installing the server, it runs npm install and afterwards it runs mvn install.


bvanwijk:
Looks good, if someone with a Unix system can confirm it works I think this can be merged.

jochem725:
This seems to be working, I agree with merging.

PR #241, opened 2016-06-15
Fix drawing subviews aspectratio

Kingdorian:
I fixed the subviews aspectratio.


jochem725:

jochem725:
I've discovered some checkstyle issues, I think these should be fixed prior to merging.

bvanwijk:
After the checkstyle warnings are fixed, I think this can be merged

NaomidR:
I've fixed the checkstyle warnings, I think this can be merged.

PR #242, opened 2016-06-15
Tag Window CSS

jochem725:
I've upgraded the tag window styling to match the edit performance window.
Please verify if tags are still added correctly. Editing a tag name is not functioning, maybe @egedebruin can take a look at this.

### DIFF
--- a/frontend/public/css/style.css
+++ b/frontend/public/css/style.css
@@ -630,12 +630,12 @@ input[type=range]:focus {
     padding: 0;
 }
 
-.performance-list span {
-    margin-right: 5px;
+.fill-tags .tag {
+    width: 100%;
 }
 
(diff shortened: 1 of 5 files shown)


egedebruin:
The tags are now updated correctly.
It looks good and I think this can be merged.

jochem725:
It works fine now, thanks Ege.

bvanwijk:
There is a style error with the lettertype in the tags window

jochem725:
I've resolved the issues, please confirm the text is formatted correctly now.

bvanwijk:
There is a bug when you click edit name in the performance window twice.

jochem725:
I've fixed the issue you were pointing out @bvanwijk, now everything should function correctly.

PR #243, opened 2016-06-16
Fix adding IPpresets to database

egedebruin:
The database did not add the IPCameraPresets to the database, this is fixed now.

### DIFF
--- a/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
+++ b/backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
@@ -276,7 +276,7 @@ public class MySQLDatabase implements Database {
       statement.setInt(9, autoir);
       statement.setInt(10, preset.getId());
     } catch (SQLException e) {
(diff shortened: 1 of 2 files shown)


bvanwijk:

backend/src/main/java/com/benine/backend/database/MySQLDatabase.java
```diff
@@ -273,7 +276,7 @@ private void setIpPreset(IPCameraPreset preset, PreparedStatement statement) {
       statement.setInt(9, autoir);
       statement.setInt(10, preset.getId());
     } catch (SQLException e) {
-      logger.log("Presets could not be added.", LogEvent.Type.CRITICAL);
+      logger.log("Presets could not be added IP.", LogEvent.Type.CRITICAL);
```
NaomidR: I don't think this is a correct sentence

NaomidR:
I've placed a small comment. After you have changed this I agree with merging.

jochem725:
I agree with merging.

PR #244, opened 2016-06-16
Load the focus and iris controls

bvanwijk:
The focus and iris controls are now loaded correctly


egedebruin:

jochem725:
I'm happy to see this fixed, I'll merge it.

PR #245, opened 2016-06-16
Separate concerns

bvanwijk:
The preset class still has a bit too much incomming calls, which needs to be fixed.


jochem725:
Looks fine, I think you can merge this.

NaomidR:
Looks really good! The only thing that I noticed is that you've got a loooot of success= false in the exceptions, maybe it would be better to change it around everywhere and start with boolean success = false.

jochem725:
I don't think I have any concerns the new changes, I agree with merging.

PR #249, opened 2016-06-16
Auto Preset Tab Menu

jochem725:
You can now do the auto preset generation with a nice GUI wizard.
Please review this extra carefully because it has not been tested with a Panasonic camera yet.

I've added a screenshot of the final preset select window because this cannot be made visible without a camera.

schermafbeelding 2016-06-16 om 18 11 53


bvanwijk:
Looks very good, I have tested the things which are possible withouth the panasonic cameras.

egedebruin:
I also think this looks very good!

Kingdorian:
The tags field of the modal resizes in a weird way for me. It has a fixed size but after you enter one or 2 characters it becomes smaller and then grows with the amount of characters you add.

backend/configs/main.conf
```diff
@@ -4,13 +4,17 @@ IPCameraTimeOut=2000 #time the values are saved of an ip camera
...
+
 camera_2_type =simplecamera
 camera_2_address=http://131.180.123.51/zm/cgi-bin/nph-zms?mode=jpeg&monitor=2&scale=100&buffer=100
 camera_2_macaddress=simplecamera2
+camera_2_aspectratio=1.7777
 
-camera_3_type=simplecamera
-camera_3_address=http://delficam.ewi.tudelft.nl/axis-cgi/mjpg/video.cgi?camera=1&resolution=704x576&1464863399284
-camera_3_macaddress=simplecamera3
+#camera_3_type=simplecamera
```
Kingdorian: I don't think we want to merge this commented version of the "default" (which is not officially default) version into develop.

jochem725: I've resolved this now.

jochem725:
@Kingdorian Thanks for pointing the input box length issue out. I've resolved it.

Kingdorian:
Since the issues I pointed out are fixed I think we can merge this, also a really nice cleanup of the css file!

jochem725:
@Kingdorian I've resolved some merge conflicts, can you confirm this has been solved correctly and then merge it?

PR #250, opened 2016-06-16
Implement show subviews created

Kingdorian:
The canvas now shows the subviews in green when presets have been created for them.


Kingdorian:
Should not be merged into master.

PR #251, opened 2016-06-16
Implement show subviews created

Kingdorian:
The canvas now shows the subviews in green when presets have been created for them.


jochem725:
I'm not sure where exactly the green color is applied, since I can only see red color codes in the code. How is the green color implemented?

jochem725:
@Kingdorian No problem, I agree with merging now.

bvanwijk:
Looks good, I agree with merging too.

PR #252, opened 2016-06-16
Bower dependencies in separate file

jochem725:
I've moved the Bower dependencies to a `bower.json` file, where the dependencies belong originally. Please verify if `npm install` is still working correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Build the back-end server using `maven install` with the supplied `pom.xml` file
 
 ### Deployable zip
 #####Unix systems:
-Run the `install.sh` script, this generates a zip file, extract this zip file to the desired location on your system. The servers can then be started by executing the `install.sh` script.
+Run the `install.sh` script, this generates a zip file, extract this zip file to the desired location on your system. The servers can then be started by executing the `startserver.sh` script. If maven is not installed on your system, run the `other systems` install instructions. 
+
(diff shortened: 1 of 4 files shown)


bvanwijk:
npm install is still working correctly, I agree with merging.

Kingdorian:
There is not much to say about this, I think this can be merged.

README.md
```diff
@@ -18,7 +18,8 @@ Build the back-end server using `maven install` with the supplied `pom.xml` file
 
 ### Deployable zip
 #####Unix systems:
-Run the `install.sh` script, this generates a zip file, extract this zip file to the desired location on your system. The servers can then be started by executing the `install.sh` script.
+Run the `install.sh` script, this generates a zip file, extract this zip file to the desired location on your system. The servers can then be started by executing the `startserver.sh` script. If maven is not installed on your system, run the `other systems` install instructions. 
+
```
Kingdorian: If maven is not installed the other systems won't work either right?

jochem725: Oh I was mistaken when editing this description. 
If we distribute a zip in the release, the install instructions are not even needed. If it is zipped correctly it should run out of the box.

Kingdorian:
The readme looks better now, I think we can merge this if the others agree.

NaomidR:
It is still working. I agree with merging.

PR #253, opened 2016-06-16
Camera tag

bvanwijk:
In the UI for every preset there is a tag "camera <id>", this makes it possible to search for presets on camera id.
I have also checked all the javascript code with jshint and let the UI only display presets which belong to an available camera.


jochem725:
It works fine, however the Camera tags are not displayed in the suggestions window. After this is fixed I agree with merging.

bvanwijk:
@jochem725 Nice find, I've fixed it.

Kingdorian:
Great enhancement I think we can merge this.

PR #254, opened 2016-06-16
Deleting cameras fix

egedebruin:
The cameras will now certainly not be removed from the database and I removed an unused method.


jochem725:
I think this is indeed better because this prevents loss of presets if a camera is deleted.

bvanwijk:
It seems to be working correctly, I agree with merging.

Kingdorian:
I think we can merge this.

PR #255, opened 2016-06-16
Fix wrong server status determination

jochem725:
I think this provides a fix for issues with server status determination. It prevents the server from hanging.


bvanwijk:
Looks like a good fix, I agree with merging.

Kingdorian:
Great fix I don't have to restart the nodejs server anymore when I restart the backend.

Issue #246, opened 2016-06-16
AutoPreset name saving is incorrect


jochem725:
It currently does not save the name entered in the modal.

Kingdorian:
Fixed in https://github.com/Kingdorian/Contextproject-BeNine/pull/256

PR #256, opened 2016-06-17
Fix names auto created presets

Kingdorian:
This fixes not adding names for automatically created presets it also fixes a bug in responding to autopresetcreation requests where the server would always return succes false in the finally block.
Fixes: https://github.com/Kingdorian/Contextproject-BeNine/issues/246

### DIFF
--- a/backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
+++ b/backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
@@ -63,6 +63,7 @@ public class AutoPresetCreationHandler extends AutoPresetHandler  {
       JSONArray idsJson = new JSONArray();
       for (int i = 0; i < presets.size(); i++) {
         presets.get(i).setName(name + i);
+        getPresetController().updatePreset(presets.get(i));
       }
       
       presets.forEach(preset -> idsJson.add(preset.getId()));
--- a/backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
+++ b/backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
@@ -14,6 +14,7 @@ import org.json.simple.JSONObject;
 
 import java.io.IOException;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


NaomidR:

backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
```diff
@@ -54,10 +55,15 @@ public void handle(String s, Request request, HttpServletRequest httpServletRequ
...
-      Collection<IPCameraPreset> presets = creator.createPresets(ipcam, creator.generateSubViews());
-      
+      ArrayList<IPCameraPreset> presets =
+              new ArrayList<>(creator.createPresets(ipcam, creator.generateSubViews()));
+
+      String name = request.getParameter("name");
       JSONObject jsonObject = new JSONObject();
       JSONArray idsJson = new JSONArray();
+      for (int i = 0; i < presets.size(); i++) {
+        presets.get(i).setName(name + i);
```
egedebruin: The preset names will not be updated in the database, you need to call the updatePreset method.

backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
```diff
@@ -66,12 +72,13 @@ public void handle(String s, Request request, HttpServletRequest httpServletRequ
     } catch (CameraConnectionException | InterruptedException
             | TimeoutException | StreamNotAvailableException | SQLException e ) {
       getLogger().log("Exception occured while trying to auto create presets", e);
+      respond(request, httpServletResponse, succes);
```
jochem725: I don't think you should return success if an exception occurs.

backend/src/main/java/com/benine/backend/http/AutoPresetCreationHandler.java
Kingdorian: Great find!

bvanwijk:

jochem725:
I see you fixed @egedebruin's issue, I think this can be merged.

egedebruin:
Good to see you have already fixed it, I agree with merging.

PR #258, opened 2016-06-17
Remove cancel button

### LINKED ISSUE #257: Cancel and close button
In the tag modal there is a cancel and close button which have the exact same functionality. One button needs to be removed.

egedebruin:
This fixed #257, I have removed the cancel button from the edit tag modal, because it was redundant.


Kingdorian:
The button was indeed redunant, I think we can merge this.

NaomidR:
Nice find, I agree with merging this.

bvanwijk:
I will merge it.

PR #259, opened 2016-06-17
Performance Ok preset creation fix

jochem725:
The performance ok button should now no longer create a preset.


Kingdorian:

bvanwijk:
Seems like a good fix.

NaomidR:
This fixes it. I agree with merging.

PR #260, opened 2016-06-17
Remove wrong responsd call in auto preset status handler

bvanwijk:
The auto preset status response is now correct.


Kingdorian:
Seems to work, I think we can merge this.

NaomidR:
Good fix, I also agree with merging.

jochem725:
I'll merge it then.

PR #261, opened 2016-06-17
Add comments to the toJSON method.

NaomidR:
This PR adds some comments to the JSON method because these were missing/ the method was too long.


jochem725:
These comments make the method more clear, I think this can be merged.

PR #262, opened 2016-06-17
Fix preset loading in performances

bvanwijk:
The presets are loaded again in the performances


NaomidR:

jochem725:
Great this is finally fixed, I was unable to find the bug earlier today. I think this can be merged.

PR #263, opened 2016-06-17
Fix performance deletion

jochem725:
I've fixed performance deletion and solved a css issue. Please confirm this is working.


egedebruin:
I have tested it and it looks fine.

bvanwijk:
Looks nice, the delete buttons are now displayed right.

NaomidR:
Looks good indeed. I also agree with merging.

PR #264, opened 2016-06-17
The focus and iris toggle are now showing the right status

bvanwijk:
The toggle buttons are now showing the right status of the iris and focus


jochem725:
I can confirm this is working. This can be merged.

Kingdorian:
Nice fix. I think we can merge this.

PR #267, opened 2016-06-17
Fix subview drawingorder

Kingdorian:
Now the higher layer subviews are drawn on top of the lower layer subviews, also each layer is slightly more inset then the previous layer.


jochem725:
I've just reviewed this code with the camera before the pull request was opened. I can confirm the subviews are now now drawn correctly. I agree with merging.

bvanwijk:
Looks really cool now, I agree with merging.

egedebruin:
The code looks fine, so if t works correctly with a camera I agree with merging.

PR #269, opened 2016-06-17
Increase version number in pom and deployable

Kingdorian:
I increased the versionnumber in the pom.xml and install and startserver scripts, please verify this works correctly.


egedebruin:
The deployable works good for me, so I think this can be merged.

bvanwijk:
Looks good, I will merge it.

PR #270, opened 2016-06-17
Fix last checkstyle warnings

bvanwijk:
Fix last checkstyle warnings


Kingdorian:
I think this can be merged.

PR #271, opened 2016-06-17
Final release

Kingdorian:
This is the pull request for the final version of the product.


bvanwijk:
I think we can merge this, since we all agreed on this in separate pull requests and everything seems to work.

egedebruin:
This looks like a great final version!
So this can go to master

PR #273, opened 2016-06-21
Fix Success typo

### LINKED ISSUE #272: Typo in json return
"Success" should be written with double s at the end
see for example the return json of presets/recallpreset

jochem725:
This should fix #272.


bryanvanwijk:
Looks like a good change

NaomidR:
Indeed a good change.

PR #274, opened 2016-06-21
Add type documentaton to config

jochem725:
I've improved the documentation in the config, so someone not involved with the project can actually use it.


bryanvanwijk:
Looks like a good change

NaomidR:
The config file looks better now.

PR #277, opened 2016-06-22
Add address and camera type to /camera response

jochem725:
By request of team goto-fail, I've added the camera type and camera address to the `/camera` response.


jochem725:
I've now added a distinction between the ip address and the stream address. For `simplecamera` types, `address` is equal to the `streamaddress` by design. For `ipcamera` types (the Panasonic camera) `address` should output the ip address of the camera, while `streamaddress` outputs the address of the motion jpeg stream.

martijnjanssen:
Nice, this is exactly what we needed!

bryanvanwijk:
Looks great, I agree with merging.

PR #278, opened 2016-06-22
Add file to create database in deployable

bryanvanwijk:
Add file to create database in deployable


jochem725:
This solves my issues with the database, thanks @bvanwijk

PR #279, opened 2016-06-23
Fix stream compatibility

jochem725:
This resolves a Panasonic camera compatibility issue.


bryanvanwijk:

egedebruin:
LGTM

